PR #4, opened 2017-07-31
Remove resources from koin-android and koin-android-support

hvisser:
Resources like `app_name` are not needed and can conflict with app resources.


arnaudgiuliani:
Thanks for your resource cleaning

PR #10, opened 2017-10-27
weatherAppModules() instead of allModules()

Jeevuz:
Changed allModules() function name to the weatherAppModules() from sample to make it more clear. Repo has no allModules() function now.


arnaudgiuliani:
yep

PR #14, opened 2017-11-02
Add new method getPropertyOrElse

Baresse:


arnaudgiuliani:
I put that in 0.5.2

Jeevuz:
Hi @Baresse, @arnaudgiuliani! 
Wonder why you need this methods?
Can't imagine when we can want to to made not provided dependency or property and get default value? I think it smells, isn't it?

arnaudgiuliani:
Hello,

Koin properties can help replace Android property handling and avoid things such as Parcelable. In this case, you pass properties to next views and don't need any default value as you pointed here. But you can also save data (like savedInstanceState stuff), but in a more straight way: just set a data that will be retrieve on rotation or something like that. In such case, the first property state need an empty/default one.

The default value for properties is not the main way to go, but rather a tool support for some annoying situations :)

Jeevuz:
Thanks for the answer. Agree for the property. But I saw in the PR changes also getOrElse(). Is it needed?

Jeevuz:
After a minute of thinking I still not really sure that the library has to handle this. Really, this is super easy to get with Kotlin. getProperty<>() ?: defaultValue. 
The main reason to not have helper methods like that is to retain the library small and focused. This is good for understanding and maintain.

arnaudgiuliani:
I've improved it & rewrite it to be just an overload of existing `getProperty(key: String)`  with `fun getProperty(key: String, defaultValue: T)`

This way we don't introduce too much things as you pointed.

Baresse:
@Jeevuz sorry for the late answer I didn't notice the mail notification on this topic. When you say : 
>Really, this is super easy to get with Kotlin. getProperty<>() ?: defaultValue.

This is unfortunately not right since `getProperty<>()` does not return a null value but throw an exception. As a consequence, should you have to test if a property is already set you had to write something less clean such as : 
```
try { 
  getProperty(key) 
} catch (e: MissingPropertyException) {  }

Jeevuz:
Hi, @Baresse!
Yes, you are right. I was misleaded by the javadoc on the left side of the PR diff where was "can return null". Thanks for pointing this out.

About naming: the `getProperty()` not corresponds to the `get()` from Map API too. `get()` returns null if key not found. `getProperty()` throws exception. 

For me is easier to change the `getProperty()` to returning `null` and use `getProperty ?: defaultValue` in code if no property was set. Returning null will say clearly "there is no property set for this key", just as map.get() do.

PR #16, opened 2017-11-08
Add feature to bind all system properties and koin properties

Baresse:
Add the capability to load properties defined in `System.Properties()` and from `koin.properties` file if found in classpath. 

Automatic system properties binding can be disabled like this : 
```
startContext(modulesList, false)
```
but by default the properties will be bond into the context


Jeevuz:
Hi @Baresse!
Why you want the library to have this?

IMHO, the library needs to stay small and focused, and all that can be easily done by hands in one place of your app have not to be included in it to save the size and the main mission of the library.

Baresse:
Hi @Jeevuz 
Exactly for the reason you are pointing out. 

The changes are really small and still focused on the library need for the properties part of the library. You will see in near future an article explaining why Koin properties are so important especially for Android developers. 

Why should any Koin users have to write by hands on every projects this kind of snippet ? If you have a look at SpringBoot or micro frameworks such as Spark Java, Ktor, etc. ; everything was not that hard to write by hands, but the real magic is there is already plenty of things right out-of-the-box. 

I agree with you Koin should not include too much things, but should provide as much facilities as possible around its two main features which are : 

 * Bean injection (with a given and well mastered life-cycle) intended for Logic or services
 * Property binding (with a more "static" life-cycle) intended for DTO or data

Jeevuz:
For me the magic of Koin is that this is the smallest and easiest library, not all-out-of-the-box library. In fact the main reason I switched to this from dagger2 is that dagger adding more and more useless for me features. I know many devs who really need only simple injection from а DI-framework. Properties is really nice addition, it solves some problems (don't need to create separate models (in MV* meaning) for passing data or assisted injection without factories or separate scopes).
My point that many developers so tired from this all-in-one DI-frameworks, that they want something simple and focused. Me either. That is why I asking why to add this to the library.
I agree that property binding is one of main features of the library, but disagree that binding concrete properties is related to this functionality.

arnaudgiuliani:
@Baresse check the 0.6.0 branch

@Jeevuz Binding properties is one of the last mission of Koin - we have several way to do it (system, from external files or directly from dsl) 
The core system must stay simple, but properties just can be only written in DSL :/ 

Stay tunned

Baresse:
@arnaudgiuliani 
I did a rebase of all the commits and change my mind for default value for SystemPropertiesBinding. 
It's false by default. This way, nothing will be change for current users of Koin library.

Baresse:
@arnaudgiuliani after the last commit the following chain is setup : 
 "=>"  means "can be overridden by"
 DSL props  => startContext given props => koin.properties props => System props

arnaudgiuliani:
Ok. I put in 0.6.0

PR #25, opened 2017-12-02
Proper call for extension fun Application.startKoin

BukT0p:
Double pass of Application reference. From Java code this method looks ugly:
AndroidExtKt.startKoin(this, this, modules, properties);

Of course this fix will brake existing samples


saied89:
Hi
Can some one give an example for how 

> modules

 is built?

PR #27, opened 2017-12-04
Fix problem in concurrent call

jxw1102:
Hi. Koin is a very nice framework, I like it very much!
When I use Koin with coroutine, the code runs into `throw IllegalStateException("Calling HEAD was $head but should be $clazz")` in `KoinContext.kt`.

My code which uses Koin is:
```
class ExampleViewModel : ViewModel(), KoinComponent {

    val product = MutableLiveData<Product>()

    val db by inject<Database>()
    val service by inject<ProductService>()

    fun requestProduct(id: Int) {
        launch(UI) {
            val productResult = service.getProduct(id)
                    .subscribeOn(Schedulers.io())
                    .observeOn(AndroidSchedulers.mainThread())
                    .await()
            product.value = productResult.get()
        }
    }

    fun getProductFromDb(id: Int) = start {
        product.value = async { db.productDao().getProduct(id) }.await()
    }

}
```
In my Activity.onCreate(), I call `viewModel.getProductFromDb(100); viewModel.requestProduct(100)`.

In order to well understand the problem, I have added one line of log in KoinContext.resolveInstance(), `logger.log("thread: ${Thread.currentThread()}, head: ${head.java.simpleName}, clazz: ${clazz.java.simpleName}")`.

Here's the log:
```
12-04 22:54:18.050 4766-4766/com.mycompany D/KOIN_LOGGER: Resolve [com.mycompany.ProductService] <> []
12-04 22:54:18.050 4766-4874/com.mycompany D/KOIN_LOGGER: Resolve [com.mycompany.Database] <> []
12-04 22:54:18.050 4766-4874/com.mycompany D/KOIN_LOGGER: thread: Thread[ForkJoinPool.commonPool-worker-2,5,main], head: ProductService, clazz: Database
12-04 22:54:18.050 4766-4766/com.mycompany D/KOIN_LOGGER: thread: Thread[main,5,main], head: Database, clazz: ProductService
12-04 22:54:18.050 4766-4766/com.mycompany E/KOIN_LOGGER: Call stack error -- []
12-04 22:54:18.051 4766-4874/com.mycompany E/KOIN_LOGGER: Call stack error -- []
```
We can see that the funciton `resolveInstance()` is called in the same time on two different threads. Then it throws the IllegalStateException.

To fix this, we can simply add a lock for this funciton's body. Tested in my case, it works.


ChristopheCiambella:
+1
I have same issue with coroutine and this code work very well in my project, thanks !

arnaudgiuliani:
Ok to integrate it. I will publish alpha version to allow you to test

arnaudgiuliani:
You can check the 0.7.0-alpha-1 version

arnaudgiuliani:
be aware that this version brings new dsl - check test sourcecode

PR #32, opened 2017-12-21
Android Arch Lifecycle management of Koin Context.

TonicArtos:
A lifecycle observer that can release the Koin Context and a couple of extension functions to make it easier to use.

----

I thought I'd just quickly crib this together from one of my projects and push it out here, in case it is useful for you or anybody following the project. The idea is to use the new Android Lifecycle Components to release the Koin context, retaining the flexibility of the existing approach. The only trouble is that it adds further dependencies to the project plus the use of the annotations processor. 

Usage
```kotlin
class MainActivity : FragmentActivity() {
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)


        manageContext("MainActivityContext")


    }
}
```

`manageContext` can be called with an optional parameter to set which lifecycle event releases the context.

Please feel free to let me know what you think.

### DIFF
--- a/android/koin-android/build.gradle
+++ b/android/koin-android/build.gradle
@@ -1,5 +1,6 @@
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-kapt'
 
 apply plugin: 'com.github.dcendents.android-maven'
 apply plugin: 'com.jfrog.bintray'
@@ -36,6 +37,9 @@ dependencies {
     })
     // Kotlin
     compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    // Android Lifecycles
+    compile "android.arch.lifecycle:runtime:$android_arch_lifecycle_version"
+    kapt "android.arch.lifecycle:compiler:$android_arch_lifecycle_compiler_version"
     // Koin
     compile "org.koin:koin-core:$koin_version"
     testCompile "org.koin:koin-test:$koin_version"
--- a/android/koin-android/src/main/java/org/koin/android/lifecycle/Context.kt
+++ b/android/koin-android/src/main/java/org/koin/android/lifecycle/Context.kt
@@ -1,0 +1,57 @@
+package org.koin.android.lifecycle
+
+import android.arch.lifecycle.Lifecycle
+import android.arch.lifecycle.LifecycleObserver
+import android.arch.lifecycle.OnLifecycleEvent
+import android.support.v4.app.Fragment
+import android.support.v4.app.FragmentActivity
+import org.koin.standalone.KoinComponent
+import org.koin.standalone.releaseContext
+
+sealed class LifecycleEvent {
+    object OnDestroy : LifecycleEvent()
+    object OnStop : LifecycleEvent()
+    object OnPause : LifecycleEvent()
+}
+
+fun FragmentActivity.manageContext(
+        contextName: String,
+        releaseEvent: LifecycleEvent = LifecycleEvent.OnDestroy
+): LifecycleObserver = LifecycleContextBinder(contextName, releaseEvent, lifecycle)
+
+fun Fragment.manageContext(
+        contextName: String,
+        releaseEvent: LifecycleEvent = LifecycleEvent.OnDestroy
+): LifecycleObserver = LifecycleContextBinder(contextName, releaseEvent, lifecycle)
+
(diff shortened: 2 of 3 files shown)


arnaudgiuliani:
Hello,

the `koin-android` module address only main Android features. I try to keep it as simple as possible. I'm thinking about a dedicated module for Android architecture components.

TonicArtos:
Okay, that sounds reasonable. Thanks for letting me know.

arnaudgiuliani:
Yeah,  I keep it opened. I will tell you how it can be included in one of the next releases.

arnaudgiuliani:
Hello,

you can check the last release (0.8.0), about bringing a ViewModel factory to Android architecture API

TonicArtos:
I'll do that in a few days. Just hopping in and out of trains and planes at
the moment.

TonicArtos:
Hi. I like where *v0.8.0* is going with adding an API for view model injection. However, there are some things I noted.

1. I cannot inject the view model in the normal manner 
    ```kotlin 
    private val viewModel: MyViewModel by inject()
    ```

2. I have to manually pull the activity or parent fragment in order to use shared view models.
    ```kotlin
    val viewModel = activity.getViewModel<MyViewModel>()
    ```
    For an activity, it is straightforward. However, for a parent fragment it is more complicated.

3. Configuring the koin context with `viewModel {}` makes it look like nested contexts should inject/get the same instance of the view model as the parent context. This could be confusing to people who have not read the underlying code.

---

Presently, I have a different approach which hides the ViewModelProvider behind the koin context and allows for lazy injection.

This approach does require the context to be initiated in Fragment::onCreate or Activity::onCreate. Also, the ViewModels are KoinComponents eliminating constructor injection and negating the need for providing a custom ViewModelFactory.

```kotlin
/**
 * Main activity using view model injection with initialisation of koin context in onCreate.
 */
class MainActivity: FragmentActivity() {
    private val navigation: NavigationViewModel by inject()

    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)

        setProperty(Context.Activity, this)
    }
}

/**
 * Home fragment injecting navigation view model which is shared with activity.
 */
class HomeFragment: Fragment() {
    private val navigation: NavigationViewModel by inject()
    private val viewModel: HomeViewModel by inject()

    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)

        setProperty(Context.HomeFragment, this)
    }
}

/**
 * Context string constants.
 */
object Context {
    const val Activity = "Activity"
    const val HomeFragment = "HomeFragment"
}

/**
 * Get list of app modules for startKoin.
 */
fun appModules() = listOf(appModule)

/**
 * Koin application module.
 */
private val appModule = applicationContext {
    context(Activity) {
        viewModel<NavigationViewModel>()

        context(HomeFragment) {
            viewModel<HomeViewModel>()
        }
    }
}

/**
 * Add a view model to the context.
 * 
 * @param scopeType Scope type of ViewModelProvider. If not set an attempt will be made to determine it at runtime.
 * @param name Name of property scope object is set within. Default is same as context.
 */
private inline fun <reified T : ViewModel> Context.viewModel(
        scopeType: ScopeType? = null,
        name: String = this.name
) = when (name) {
    Scope.ROOT -> throw CantInjectViewModelException()
    else       -> factory {
        when (scopeType) {
            ScopeType.Activity -> ViewModelProviders.of(getActivity()).get(T::class.java)
            ScopeType.Fragment -> ViewModelProviders.of(getFragment()).get(T::class.java)
            null               -> try {
                ViewModelProviders.of(getActivity()).get(T::class.java)
            } catch (e: ClassCastException) {
                try {
                    ViewModelProviders.of(getFragment()).get(T::class.java)
                } catch (e: ClassCastException) {
                    throw CantInjectViewModelException()
                }
            }
        }
    }
}

sealed class ScopeType {
    object Activity : ScopeType()
    object Fragment : ScopeType()
}

private class CantInjectViewModelException :
        IllegalStateException("Injecting view model requires associated Fragment or FragmentActivity.")

private inline fun Context.getActivity(name: String = this.name) = getProperty<FragmentActivity>(name)
private inline fun Context.getFragment(name: String = this.name) = getProperty<Fragment>(name)
```

arnaudgiuliani:
Hello I propose to read incoming changes from koin-android-architecture module branch master. Feel free to give any suggestions. I close this PR then.

PR #33, opened 2018-01-04
Koin android architecture lazy inject

fredy-mederos:
This request is to archive the following feature:
```
// Bind it to your Activity
val viewModel by injectViewModel<MyViewModel>()

class MyActivity() : AppCompatActivity() {
  override fun onCreate(){
    super.onCreate()
    viewModel.someFunction()
  }
}
```

### DIFF
--- a/android/koin-android-architecture/src/main/java/org/koin/android/architecture/ext/AndroidExt.kt
+++ b/android/koin-android-architecture/src/main/java/org/koin/android/architecture/ext/AndroidExt.kt
@@ -10,12 +10,12 @@ import org.koin.standalone.KoinComponent
 
 /**
  * Retrieve given ViewModel for Android Fragment
- * @param shareWithActivityViewModel - if the viewModel instance is shared between activity/fragment / optional (false by default)
+ * @param shareWithActivityViewModel - if the viewModel instance is shared between activity/fragment / optional (true by default)
  */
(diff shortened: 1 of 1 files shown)


arnaudgiuliani:
I would add some improvements on how write the lazy delegate, directly reuse the getViewModel() function

fredy-mederos:
You are right! I did it in my project but I forget to commit it in the fork

arnaudgiuliani:
I will provide an update for lazy get viewModel and fragment integration. Once available on 0.8.1 I will notice you to check it.

arnaudgiuliani:
Not sure if you reuse the same ViewModel with several fragments, you intend to reuse the same instance. Else you use separated ViewModel. This makes things simple to understand.

For now, I think to propose shared viewmodel only. We will see if there's need for recreating the same viewmodel instance for fragment

fredy-mederos:
So I can remove the `shareWithActivityViewModel: Boolean` param and always use the `ViewModelStores.of(activity)` to use the shared viewmodel only

arnaudgiuliani:
I suggest a default parameter to always get VM from activity, and allow to create a new one if needed.

I will merge your code and notify you for an alpha version of koin to check.

Stay tunned!

fredy-mederos:
Ok thanks!

arnaudgiuliani:
Check the 0.8.1-alpha-7 👍

Issue #43, opened 2018-01-30
java.lang.NoClassDefFoundError: java.util.concurrent.ConcurrentHashMap$KeySetView


murselturk:
Hi,

This is a great library. However I am getting the following error after the update to 0.8.1, if I run in a device with SDK 16:

`E/AndroidRuntime: FATAL EXCEPTION: main
                                                                     java.lang.NoClassDefFoundError: java.util.concurrent.ConcurrentHashMap$KeySetView
                                                                         at org.koin.core.instance.InstanceFactory.findInstance(InstanceFactory.kt:45)
                                                                         at org.koin.core.instance.InstanceFactory.retrieveInstance(InstanceFactory.kt:27)
                                                                         at org.koin.KoinContext.resolveInstance(KoinContext.kt:70)
                                                                         at com.murselturk.radyo7.ui.base.BaseActivity$$special$$inlined$inject$1.invoke(AndroidExt.kt:106)
                                                                         at kotlin.SynchronizedLazyImpl.getValue(Lazy.kt:131)`

Any ideas what to do?
Thanks.

arnaudgiuliani:
Hello,

the upgrade of Kotlin in `1.2.21` effectively broke the API of our instance registry and make it like it were in jdk 8.

I'll fix it for 0.8.2 👍

m-ivanov:
In android in broken below API level 24

murselturk:
I've also test in a device with SDK 22 and had no error messages. Maybe below API Level 21

m-ivanov:
i desumed 21 from KeySetView  android docs, but tested on Api 19 emulator with the same crash, i it can help i can provide my gradle

arnaudgiuliani:
Can you try the `0.8.2-alpha-14`?

murselturk:
Works like a charm now :)

arnaudgiuliani:
Nice. I push it to `0.8.2` thanks !

arnaudgiuliani:
Don't hesitate if you see problem like this one 👍

richardroche:
Thank you so much for being so proactive

PR #44, opened 2018-01-31
ConcurrentHashMap.KeySet()

murselturk:
Solving compatibility issue with ConcurrentHashMap.KeySet()

http://valery-frolov.blogspot.de/2015/04/javalangnosuchmethoderror.html

Issue: #43


arnaudgiuliani:
Nice, thanks for your help.

I'm trying another solution of throwing away this ConcurrentHashMap ... sync is made at higher level and then no need of concurrenthashmap. But I keep your solution if I have any problem.

arnaudgiuliani:
I close your PR. It should be ok with the fix from `0.8.2`

Thanks for your feedback !

PR #52, opened 2018-02-15
Documentation fixes

lesraf:
During Koin integration I stumbled upon things below.
Changed _MyViewModel_ val name to _myViewModel_, because it was interfering with _by viewModel()_ function name and it took me some time to find out what's going on because errors weren't that helpful.


arnaudgiuliani:
Ok, thanks. I'll update the website too.

PR #58, opened 2018-02-18
Allow retrieving ViewModels from stores by key

amaksoft:
Sometimes it is necessary to store multiple instances of same `ViewModel` type in the one `ViewModelStore`. The only way to resolve this case is to call `ViewModelProvider.get(String key, Class<T> modelClass)` with a String key parameter. I think we need to add a key parameter to `viewmodel` extension if possible.


arnaudgiuliani:
Hello,

Ok I see. Do you have a concrete example on using different instances of the same ViewModel?

amaksoft:
Sometimes I use generic ViewModels in my projects. Here is an [example](https://gist.github.com/amaksoft/32099764efb7f4762c3832c40ff21ef1).
It's a generalized ViewModel for making an HTTP request without creating similar classes for different screens/requests.

Here is how to use it:

```kotlin
val wiewModel = ViewModelProviders.of(fragment).get("UpdateUser", RequestViewModel.class) as RequestViewModel<User>
viewModel.response.observe(this, Observer { resp: Response<User> ->
    showProgress(resp?.status == Response.Status.LOADING)
        when (resp?.status) {
            Response.Status.SUCCESS -> {
                updateUserFields(resp?.data)
            }
            Response.Status.ERROR -> {
                showError(resp?.error)
            }
            Response.Status.LOADING -> {}
        }
})

btSave.setOnCLickListener {
    viewmodel.makeRequest(httpService.updateUser(makeUpdatedUser()))
}

```

This way you can have dozens of view models of the same type and even same type parameter, but making different requests. I'm not saying it's the best practices, just exploring this pattern. Feel free to use it if you like.
It's merely an example. I'm sure people can think of more ways to use multiple instances of ViewModel of the same type.

arnaudgiuliani:
Yes I understand. I will make the patch on a next `rc` version. Can you check it? I will provide it in `0.9.0-rc-3`

Can be cool if you have a very small project example to test this `key` feature.

amaksoft:
Sure! I can even add a robolectric based unit-test to my PR to test this

arnaudgiuliani:
I've just redirect your PR on the right branch. Check the conflicting diffs.

Issue #75, opened 2018-03-10
Inject generic type?


caleb-allen:
Is it possible to inject a generic type? I'm getting a cyclical error when I try to inject two generic types. For example:

if I had `class Generic<T>` can I start koin like this?

    bean { Generic<FirstClass>() }
    bean { Generic<SecondClass>() }

I'm only getting cyclical errors, even when I include a dependency name. Any help would be appreciated.

    org.koin.error.DependencyResolutionException: Cyclic dependency detected while resolving my.Generic - [Factory[name='first', class=my.Generic]]

caleb-allen:
My thought is that naming the bean definition should override the class definition when resolving a bean instance

arnaudgiuliani:
As I told you in PR, if you have generic type to make as definition, just use names like:

```kotlin
val module = applicationContext {
        bean("a") { ComponentA() as InterfaceComponent<String> }
        bean("b") { ComponentB() as InterfaceComponent<Int> }
    }
```

caleb-allen:
Solved by c02e72ed723704fe78a2e98046a178b4edbe6f68

Thanks!

majkrzak:
@caleb-allen how it was solved?
In `2.0.0-beta4` there is still no possibility to inject by generic type

@arnaudgiuliani :point_up:

erickok:
@majkrzak Koin doens't support generics the way you are probably thinking of. The fix only pertained to named dependencies.

Some solution are using qualified (named) dependencies or non-generic definitions.

caleb-allen:
Just saw this. I think @erickok is right, I don't recall specifically what I was doing but I believe it was fixing an issue where `Foo<Bar>` and `Foo<Baz>` would be interpreted as the same, even with a named dependency definition and this fixed that.

magillus:
I think in 2.0 there is that `bind`/`binds` on BeanDefinition to add additional binding type.

magillus:
I did it with special inline method per generic type to handle:
https://gist.github.com/magillus/07c6b188bafd927fe9944e049e7d8343

PR #76, opened 2018-03-10
Take bean name into account when checking for cycles

caleb-allen:
Solves #75 

`KoinContext#resolveInstance` now takes into account bean names when beans are being resolved by name. This allows instances of the same type to be injected to one another, so long as they have unique names. Cycles are still prevented by climbing up the resolution stack, as before, it just takes into account the name.

This PR also prints out the stacktrace of the exception that prevents the instance creation.

This is helpful to users trying to debug what is preventing their Koin implementation from working, otherwise the entire stacktrace is swallowed and we only get the exception type.

### DIFF
--- a/koin-core/src/main/kotlin/org/koin/KoinContext.kt
+++ b/koin-core/src/main/kotlin/org/koin/KoinContext.kt
@@ -45,7 +45,8 @@ class KoinContext(
      * @param name bean definition name
      */
     inline fun <reified T> resolveByName(name: String, noinline parameters: ParameterMap): T =
-        resolveInstance(T::class, parameters) { beanRegistry.searchByName(name) }
+        resolveInstance(T::class, parameters, {bean -> (bean.clazz == T::class || bean.types.contains(T::class)) && bean.name == name})
+        { beanRegistry.searchByName(name) }
 
     /**
      * Resolve a dependency for its bean definition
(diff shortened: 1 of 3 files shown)


arnaudgiuliani:
Hello,

 I was reworking the stack resolution as a new component. Ok for the stacktrace stuff (too much code clean of stack traces :D ).

For your proposal on bean resolution with names, what is your need behind? For generic beans, just use names:

```kotlin
val module = applicationContext {
        bean("a") { ComponentA() as InterfaceComponent<String> }
        bean("b") { ComponentB() as InterfaceComponent<Int> }
    }
```

caleb-allen:
I don't think that my use case specifically is about generics, but more about how Koin detects cycles (and it happens to be because of generics). The need is for generics that may have the same type; in the way that I should be able to have multiple beans for a single type , for example

     bean("a") { Component() as InterfaceComponent }
     bean("b") { Component() as InterfaceComponent }

I should also be able to create any amount of `beans` or `factories` for a single type, whether generic or not, but more importantly those dependencies should be able to be injected as long as there are no cycles, regardless of type.

As the code exists now, let's say I have a `Config` type that is initialized based on traits of another Config object:

    class Config(val existingConfig: Config? = null) { . . . }

    val module = applicationContext {
        bean("globalConfig") { Config() }
        bean("userConfig") { Config(get("globalConfig")) }
    }

Koin currently detects cycles only by the class name, and this would throw:

    org.koin.error.DependencyResolutionException: Cyclic dependency detected while resolving my.Config- [Factory[name='userConfig', class=my.Config]]

so if I had

    class NeedsConfig(val config: Config) { }
    . . .
    bean { NeedsConfig(get("globalConfig)) }

that would work, because I'm only ever accessing 1 instance of `Config`, `Config.class` is only added to the resolution stack once, and the resolution logic never encounters another class of the same type.

When resolving a dependency by name, Koin first verifies that the dependency isn't already in the resolution stack, but only by class name and the bound types. If I want multiple dependencies of the same type the verification doesn't take into account the bean name. The _resolution_ and actual instance creation of the class does it correctly, but we don't get to that point because the resolution stack already has the class name in it.

With this PR, we pass in a lambda to `resolveInstance` which serves to verify that the existing resolution stack doesn't clash with the requested dependency, taking into account the name (and could be extended if there were other ways of identifying dependencies).

caleb-allen:
Just took a look at your recent commits, it looks like the `ResolutionStack` in c02e72ed723704fe78a2e98046a178b4edbe6f68 solved this issue! Directly comparing the bean definitions takes into account the bean name even if they are of the same type!

Cheers

PR #77, opened 2018-03-11
Add possibility to request dependency on every property call

kirich1409:
Need to have possibility to inject every time new instance of dependency. Now library only has possibility to inject dependency lazy via `inject`. My PR add functional similar to Java JSR-330 [Provider](https://docs.oracle.com/javaee/6/api/javax/inject/Provider.html).

### DIFF
--- a/android/koin-android/src/main/java/org/koin/android/ext/android/AndroidExt.kt
+++ b/android/koin-android/src/main/java/org/koin/android/ext/android/AndroidExt.kt
@@ -10,6 +10,8 @@ import org.koin.android.logger.AndroidLogger
 import org.koin.dsl.module.Module
 import org.koin.log.Logger
 import org.koin.standalone.StandAloneContext
+import kotlin.properties.ReadOnlyProperty
+import kotlin.reflect.KProperty
 
 
 /**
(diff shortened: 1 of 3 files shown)


arnaudgiuliani:
Hello,

I don't understand well the need behind that. Can you give me a complete use case to illustrate that?

`factory` declaration and dynamic parameters allow you to better shape your injection.

kirich1409:
Problem is that inject this is lazy get of dependency and it will not change until object will be destroyed. But in some cases a need to create every time new dependency instance with `factory` declaration.

Case when I need this: I have data `Repository` (_bean_) that contains functions to execute `Request` (_factory_). Every `Request` is special class that encapsulate logic of caching, error handling and other. On every request it has own dependencies. In current implementation only possibility is to create instance of `Request` inside `Repository`

arnaudgiuliani:
I think that you could use a factory provider:

`factory { create your `Request` }`

You can also use dynamic parameters to be use when injecting it:

`factory{ params -> makeRequest(params['a_param'])}`
`val request : by inject { mapOf('a_param' to <givenValue>) }`

It is a particular factory usage?

kirich1409:
Why `inject` is lazy? I don't understand this case when I started to work with Koin, that why I have problem and need simple case to work with `factory` declaration as expected. `inject` cache value no matter is this dependency declared as `bean` or `factory`.

kirich1409:
```kotlin
class TeacherRepository : KoinComponent {
    // Every call request dependency. NetworkBoundResource declaraed as factory
    private val networkBoundResource: NetworkBoundResource<List<TeacherDbo>, Items<TeacherDto>>
            by provider(NetworkBoundResources.TEACHERS)

    fun loadTeachers(force: Boolean = false): LiveData<Resource<List<TeacherDbo>>> {
        return with(networkBoundResource) {
            load(force)
            asLiveData()
        }
    }
}
```

PR #78, opened 2018-03-11
Add possibility to get dependency from Android Component

kirich1409:
Add possibility to get dependency from Android Components without inject it

### DIFF
--- a/android/koin-android/src/main/java/org/koin/android/ext/android/AndroidExt.kt
+++ b/android/koin-android/src/main/java/org/koin/android/ext/android/AndroidExt.kt
@@ -61,7 +61,7 @@ fun Application.bindBool(id: Int, key: String) {
  * inject lazily given dependency for Android component
  * @param name - bean name / optional
  */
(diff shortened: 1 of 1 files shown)


arnaudgiuliani:
Hello,

you mean to have the `get()` eager injection function into Android components. 

I push the semantic on the `by inject()` style lazy properties declaration. But why not unlock the `get()` function. But I think that most cases can be declared with lazy property injection. The `get()` function should be used to inject a dependency on a local/short function.

kirich1409:
In some case i need dependency to only execute single operation and no need a property. Simple case from Android: main Activity of app check state from preference (that is dependency in Koin) and after this start another screen based on preference value.

kirich1409:
This possibility is analogue of method injection from Dagger 2.

arnaudgiuliani:
Ok

arnaudgiuliani:
Please check conflicts

arnaudgiuliani:
I already added the fix - https://github.com/Ekito/koin/commit/789c95fba360057bcd23e6c04045295543de4c91

PR #85, opened 2018-03-17
Decreased nesting levels

elifarley:
... by leveraging natural flow interruption already provided by `return` and `throw` statements.

Having less nesting and interrupting flow early on usually makes the code easier to read.

---
This change is [Reviewable](https://reviewable.io/reviews/ekito/koin/85)


arnaudgiuliani:
Yes I understand, but's not hat I want here. 

I want to clearly identify cases of exception.

elifarley:
I see. Thanks for the explanation :)

PR #88, opened 2018-03-19
Remove application parameter from startKoin function on Android

kirich1409:


arnaudgiuliani:
Thanks but I need to deprecate the old API to push such fix, to not break things.

I close it.

kirich1409:
I made save changes. I deprecate old app when add new function

Issue #89, opened 2018-03-20
Beans with same name providing different types.


teregulovrustem:
I would like to ask if it is possible to resolve the following configuration correctly (it is synthetic example, since production code is under NDA):
```kotlin
val module = applicationContext {
    bean("global") { Page1() as Page }
    bean("global") { Book(get<Page>("global")) }

    context("local") {
        bean("local") { Page2() as Page }
        bean("local") { Book(get<Page>("local")) }
    }
}
```
When I tried to use class, which injects book:
```kotlin
class Test : KoinComponent {
    val globalBook: Book by inject("global")
    fun test() {
        println(globalBook.page.text)
    }
}
```
It failed with the following exception message:

> Multiple  definitions found to resolve type 'koin.Book' - Koin can't choose between :
	Bean[name='global', class=koin.Page]
	Bean[name='global', class=koin.Book]
	Check your modules definition or use name attribute to resolve components.

It took a moment to look through the sources (it was a pleasure to read it). KoinContext uses BeanRegistry#searchByName(name: String) and ignores requested class type information.  
I would like to ask, is this an intended behavior? If it is, then I would like to ask, is there a simple way to resolve described configuration?

If it is not, it can be changed by few lines of code.

In KoinContext:
```kotlin
inline fun <reified T> resolveByName(name: String, noinline parameters: Parameters): T = 
            resolveInstance(T::class, parameters) { beanRegistry.searchByName(name, T::class) }
```
And in BeanRegistry:
```kotlin
fun searchByName(name: String, clazz: KClass<*>): List<BeanDefinition<*>> = 
            searchDefinition { it.name == name && it.types.contains(clazz) }
```

arnaudgiuliani:
Hello,

interesting case. I think the use case has not be raised until now :)

I will add it to the `0.9.2`

teregulovrustem:
I may prepare a pull-request for it if it will be useful.
And I would like to ask about planned release date for 0.9.2?

arnaudgiuliani:
Yes open an PR.
I can release it quickly. Let's say 10 days max.

PR #92, opened 2018-03-22
Allowing to receive different types for the same bean name

teregulovrustem:
PR resolving issue #89


HKhademian:
This is not allowed in `InsertKoinIO:2.0.1` with `factory` and `single` providers.

PR #104, opened 2018-03-30
Minor cleanup

wcarmon:
Change #1: List -> Collection
I don't think order matters and using collection allows people to pass both `Sets` and `Lists`

Change #2: kotlindoc
The method isn't limited to StandaloneContext so I reworded the comment

### DIFF
--- a/koin-core/src/main/kotlin/org/koin/Koin.kt
+++ b/koin-core/src/main/kotlin/org/koin/Koin.kt
@@ -50,9 +50,9 @@ class Koin(val koinContext: KoinContext) {
     }
 
     /**
-     * load given list of module instances into current StandAlone koin context
+     * load given collection of module instances into current koin context
      */
-    fun build(modules: List<Module>): Koin {
+    fun build(modules: Collection<Module>): Koin {
         modules.forEach { module ->
             registerDefinitions(module())
         }


arnaudgiuliani:
No need to make it as collection. is there a real need behind?

wcarmon:
@General:
This may be more philosophical.  To me this is about the contract your method exposes. 

Your method doesn't care about order.  Accepting a [`List`](https://docs.oracle.com/javase/8/docs/api/java/util/List.html) implies that order is important and duplicates are acceptable.  Accepting a [`Collection`](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html) reflects the reality of your implementation (that is makes very few assumptions about the input).

It's the same reason a method might accept a [`Reader`](https://docs.oracle.com/javase/8/docs/api/java/io/Reader.html) instead of only accepting a [`FileReader`](https://docs.oracle.com/javase/8/docs/api/java/io/FileReader.html).

@Use-case:
I was trying to send a [`Set`](https://docs.oracle.com/javase/8/docs/api/java/util/Set.html).  ... so for now, it's just slightly more tedious to pass [`mySet.toList()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/to-list.html)

arnaudgiuliani:
Ok. Why don't you update the other starter functions? (in StandaloneContext and android)

arnaudgiuliani:
This is only a patch against the `Koin` class

arnaudgiuliani:
The order is important, for override rules. It's name the same to use module A & module A' or A' & A ... Last overriden definition is not the same :/

PR #123, opened 2018-05-15
Add explicit bean override functionality

svenjacobs:
I had the following code

```kotlin
import org.koin.dsl.module.applicationContext
import org.koin.standalone.KoinComponent
import org.koin.standalone.StandAloneContext.startKoin
import org.koin.standalone.inject

typealias StringProvider = () -> String
typealias IntProvider = () -> Int

class Example(private val stringProvider: StringProvider,
              private val intProvider: IntProvider) {

    fun run() {
        // Output is:
        //   1337
        //   1337
        // Output should be:
        //   Hello world
        //   1337
        println(stringProvider())
        println(intProvider())
    }
}

object Main : KoinComponent {

    private val example: Example by inject()

    fun run() {
        example.run()
    }
}

fun createModule(stringProvider: StringProvider,
                 intProvider: IntProvider) = applicationContext {

    factory { stringProvider }

    factory { intProvider }

    factory { Example(get(), get()) }
}

fun main(args: Array<String>) {
    startKoin(listOf(createModule(stringProvider = { "Hello world" },
                                  intProvider = { 1337 })))
    Main.run()
}
```

and first was wondering why the output is

```
1337
1337
```

instead of

```
Hello world
1337
```

Since the function types in this example resolve to `kotlin.jvm.functions.Function0<R>` and the generic type is lost during runtime due to JVM's type erasure I'm aware that the solution to this problem is

```kotlin
factory("stringProvider") { stringProvider }

factory("intProvider") { intProvider }

factory { Example(get("stringProvider"), get("intProvider")) }
```

However before the first factory was "silently" overridden by the second factory because both resolve to `Function0` (yes I know, the override is logged to console).

I suggest that Koin should have an option ("strict mode"?) that turns overrides into runtime exceptions so that all types must be either unique or have names. This should help in reducing unexpected behaviour.

### DIFF
--- a/koin-android-architecture/src/main/java/org/koin/android/architecture/ext/KoinExt.kt
+++ b/koin-android-architecture/src/main/java/org/koin/android/architecture/ext/KoinExt.kt
@@ -17,9 +17,10 @@ import org.koin.standalone.StandAloneContext
  */
 inline fun <reified T : ViewModel> Context.viewModel(
     name: String = "",
+    override: Boolean = false,
     noinline definition: Definition<T>
 ) {
-    val bean = factory(name, definition)
+    val bean = factory(name, override, definition)
     bean.bind(ViewModel::class)
 }
 
--- a/koin-core/src/main/kotlin/org/koin/core/bean/BeanDefinition.kt
+++ b/koin-core/src/main/kotlin/org/koin/core/bean/BeanDefinition.kt
@@ -26,6 +26,7 @@ data class BeanDefinition<out T>(
     var types: List<KClass<*>> = arrayListOf(),
     val scope: Scope = Scope.root(),
     val isSingleton: Boolean = true,
+    val override: Boolean = false,
     val definition: Definition<T>
 ) {
 
@@ -42,7 +43,7 @@ data class BeanDefinition<out T>(
      */
     fun isNotASingleton() = !isSingleton
 
-    private fun boundTypes(): String = "(" + types.map { it.java.canonicalName }.joinToString() + ")"
+    private fun boundTypes(): String = "(" + types.joinToString { it.java.canonicalName } + ")"
 
     override fun toString(): String {
         val n = if (name.isBlank()) "" else "name='$name', "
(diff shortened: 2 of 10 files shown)


svenjacobs:
@arnaudgiuliani Please see attached my try at this functionality. I've also added the necessary tests to `OverrideTest`.

However there is one issue with the last test: I thought that when a context is released the bean definitions of this context are also removed. However only the instances are dropped and the definitions are still kept in `BeanRegistry`. So when you reload a module with a sub context where an override is not explicitly allowed *after* the context has been released, you will still receive the `BeanOverrideException`. I'm not sure if this is a desired behaviour or not.

Also the override flag is `false` by default. Just to be clear: **This might break a lot of applications that implicitly used overrides before!**

arnaudgiuliani:
Hello @svenjacobs, the call of `StandAloneContext.releaseContext()` only drop instances from `InstanceFactory`. 

> So when you reload a module with a sub context where an override is not explicitly allowed after the context has been released, you will still receive the BeanOverrideException. I'm not sure if this is a desired behaviour or not.


Sorry, I don't understand. Why reload modules?

svenjacobs:
> Sorry, I don't understand. Why reload modules?

Imagine an Android environment. I want to provide the current `Context` via a module. So I have something like

```kotlin
fun createAndroidModule(private val context: Context) = applicationContext {
  context("ANDROID") {
    bean { context }
  }
}
```

This module is created and loaded in `onCreate()` of the `Activity`. The Koin context is released in `onDestroy()` so that nothing leaks. Now a second Activity should be able to "reload" this module (reinstalling it) with the new Android `Context`.

arnaudgiuliani:
available in koin `1.0.0-alpha-18`

svenjacobs:
Just curious, why did you rewrite this feature?

arnaudgiuliani:
rewrite the feature? The override was not explicit and could be done without wanting it.
Koi container will detect if you allowed overriding for your definition.

PR #148, opened 2018-06-14
Added support for dependency on all instances of certain type

amirabiri:


koin-core/src/main/kotlin/org/koin/KoinContext.kt
```diff
@@ -71,27 +79,41 @@ class KoinContext(
...
-            }
+        val clazzName = clazz.java.canonicalName
 
-            resultInstance = instance
+        val instances: MutableSet<T> = HashSet()
+
+        val beanDefinitions = getAllVisibleBeanDefinitions(clazzName, definitionResolver, resolutionStack.last())
+
+        val logIndent = resolutionStack.indent()
+        for (beanDefinition in beanDefinitions) {
```
badoualy: You should be able to use `mapNotNull` here instead (see https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/map-not-null.html )
Something like
```
return beanDefinitions.mapNotNull { resolveBeanDefinition<T>(clazzName, it, parameters, logIndent) }.toSet()
```

Or `mapNotNullTo` if you want to collect in a set directly.

badoualy: Maybe null result from `resolveBeanDefinition` shouldn't be ignore be raise an exception, as done in the "single" version of the method. Ignoring them could lead to nasty behaviors at runtime. Will need to ask @arnaudgiuliani to be sure.

kuFEAR:
I think It should be implemented with explicit binders instances which we expect to collect into set/list/map. I may declare some instance which shouldn't be part of set of all instances,

My current multi-binding approach:
```
factory("set-of-validator-binds") { mutableSetOf(ValidatorA(), ValidatorB(), ValidatorC()) }
factory<Set<Validator<Entity>>("set-of-validator", override = true) { get<MutableSet<Validator<Entity>>>("set-of-validator-binds") }
factory { MessageProcessor(get("set-of-validator"))}
```
additional binding:
```
factory("set-of-validator") { get<MutableSet<Validator<Entity>>>("set-of-validator-binds").apply { add(ValidatorE()) } }
```

It would be preferred to implement separated definition methods for multi-binding:
```
bindIntoSet<Validator<Entity>>("set-validator-a") { ValidatorA() } 
bindIntoMap<Class, Validator<Entity>>("map-validator-a") { EntityA::class to EntityValidatorA() }
bindIntoMap<Int, Validator<Entity>>("map-validator-b") { 0 to EntityValidatorB() }
bindIntoMap<String, Validator<Entity>>("map-validator-c") { "C" to EntityValidatorC() }
factory { MessageProcessor(get("set-validator-a"), get("map-validator-a"), get("map-validator-b"), get("map-validator-c")) }
```

or it may be placed inside current definition methods: 
```
factory { bindIntoMap{...} }
single { bindIntoSet{...} }
scope() { bindIntoSet{...} } 
```

arnaudgiuliani:
this PR is out of sync. You can propose something for koin 2.0 thanks.

PR #154, opened 2018-06-26
Android Jetpack support for viewmodel

yasitha999:
- Copy pasted extension functions to androidx package and updated with androidx dependencies


arnaudgiuliani:
All is already done in `koin-androidx-viewmodel` package: https://github.com/InsertKoinIO/koin/tree/1.0.0/koin-projects/koin-androidx-viewmodel

Issue #128, opened 2018-05-15
Support `Route.inject()` help in koin-ktor


ValentinTrinque:
*Note: I am a Kotlin beginner*

I am currently working on integrating an HTTP API using ktor, using koin-ktor. While designing my endpoints routing, I wanted to inject the dependencies at the `Route` level instead of `Application` level. Indeed, I split my `Route`s in seperate files, one for each resource :

**api/Application.kt**
```kotlin
fun Application.endpoints() {
    routing {
        v1()
    }
}

fun Routing.v1() {
    route("/api/v1") {
        users()
    }
}

```

**api/v1/Users.kt**
```kotlin
fun Route.users() {
    val usersQueryRepository: UsersQueryRepository by inject()

    get("/users") {
        val users: List<UserDto> = usersQueryRepository.list()

        call.respond(users)
    }
}
```

Your library implements `inject()` (& co.) helpers at `Application` level only. I was thinking about making this available at `Route` level.

Here is a working implementation:

```kotlin
inline fun <reified T> Route.inject(name: String = "") =
        kotlin.lazy { (StandAloneContext.koinContext as KoinContext).get<T>(name, emptyParameters()) }
```

What do you think ? I would be glad to help!

scottsteen:
What I would rather see is the functions defined in `KoinComponent` to not be associated with an interface. This would remove the need for the `koin-ktor` project. I plan to open another issue with this request though. Kotlin actually allows you to define functions that do not belong to a class!

To get around the limitation you face is to create a Kotlin `object` that implements the `KoinComponent` interface. This way I can call `inject()` anywhere I need to.

Example:
```
object KoinProxy : KoinComponent
...
val usersQueryRepository: UsersQueryRepository by KoinProxy.inject()
...
```

Something else to consider is to call `startKoin()` in the object's `init` block. E.g.

```
object KoinProxy : KoinComponent {
    init {
        startKoin(listOf(your_context))
    }
}
```

arnaudgiuliani:
Hello @scottsteen 

> What I would rather see is the functions defined in KoinComponent to not be associated with an interface.

This is another issue. We'll continue here about ktor Route API integration.

r03:
For me this is not usable because the inject happens at application start and not on a http request.
I would like to create a new service at every request to that route.
For now I do the inject in every get() method instead. Is there a better way?
Maybe this is more a Ktor problem then a Koin problem.

This is the extension method I use for now:
```
[...]

PR #159, opened 2018-07-06
Add Koin extensions on Routing and Route classes

Baresse:
From issue #128, I've added the Koin extension to Routing class according to KTOR route extraction mechanism (http://ktor.io/servers/structure.html#extracting-routes) and as requested in the issue to Route class also.

### DIFF
--- a/koin-projects/koin-ktor/src/main/kotlin/org/koin/ktor/ext/KtorRouteExt.kt
+++ b/koin-projects/koin-ktor/src/main/kotlin/org/koin/ktor/ext/KtorRouteExt.kt
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2017-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


koin-projects/koin-ktor/src/main/kotlin/org/koin/ktor/ext/KtorRouteExt.kt
```diff
@@ -0,0 +1,109 @@
...
+import org.koin.core.KoinContext
+import org.koin.core.parameter.ParameterDefinition
+import org.koin.core.parameter.emptyParameterDefinition
+import org.koin.standalone.StandAloneContext
+
+
+/**
+ * Ktor Koin extensions for Routing class
+ *
+ * @author Arnaud Giuliani
```
arnaudgiuliani: no need to add my name. Just yours for now

koin-projects/koin-ktor/src/main/kotlin/org/koin/ktor/ext/KtorRoutingExt.kt
```diff
@@ -0,0 +1,109 @@
...
+import org.koin.core.KoinContext
+import org.koin.core.parameter.ParameterDefinition
+import org.koin.core.parameter.emptyParameterDefinition
+import org.koin.standalone.StandAloneContext
+
+
+/**
+ * Ktor Koin extensions for Routing class
+ *
+ * @author Arnaud Giuliani
```
arnaudgiuliani: idem, no need to add my name. Just yours for now

PR #206, opened 2018-08-24
agp 3.+ is automatically uses a default version for Build Tools

satoshun:
i think, It is not necessary to specify the version of Build Tools. cuz AGP 3.+ is automatically uses a default version.

ref: https://developer.android.com/studio/releases/build-tools

### DIFF
--- a/koin-projects/examples/android-mvp/build.gradle
+++ b/koin-projects/examples/android-mvp/build.gradle
@@ -6,7 +6,6 @@ apply from: '../../gradle/versions-examples.gradle'
 
 android {
     compileSdkVersion android_target_version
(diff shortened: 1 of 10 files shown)


ychescale9:
I still prefer to specify `buildToolsVersion` explicitly as sometimes the latest version of build tools (e.g. 28.0.2) isn't used by the latest version of AGP alpha. This can cause multiple versions of build-tools being downloaded in a docker environment if the image uses a different build tools version than the one AGP chooses to use.

satoshun:
i didn't consider docker(or CI). It seems better to specify buildToolsVersion explicitly.

thx!

PR #207, opened 2018-08-25
Replace mockito-core with mockito-inline and remove MockMaker files.

ychescale9:
This PR replaces `mockito-core` with `mockito-inline` which has inline mock maker preconfigured to avoid having to include `src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker` in all projects.

This also means consumers of `koin-test` don't need to include **mockito** explicitly if not using **mockito** for mocking, which was the main motivation for this PR as I'm using **mockk** and would like to not add `mockito-inline` or the `MockMaker` file to my test source.


arnaudgiuliani:
Thanks 👍

PR #211, opened 2018-08-28
Add a Timber logger for Android projects.

ychescale9:
Adds a `TimberLogger` in a new `koin-logger-timber` project for Timber users. Please let me know if you prefer this to be added in the `koin-android` project with `compileOnly "com.jakewharton.timber:timber:$timber_version"`


erickok:
In my opinion this should not be in the core Koin project. It can easily be a seperate repo. Why add timber and not any other JVM logging framework?

PR #239, opened 2018-09-24
Update English phrasing

jschmid:
"Actual" is a _faux ami_ in french and does not mean "actuel". A better wording would be "Current".


arnaudgiuliani:
Merci ;)

PR #247, opened 2018-09-28
updated gradle version to be able open the samples in Android Studio 3.2

Dmitry-Borodin:
and removed unnecessary imports for jar files - we don't use them and it is a good practice to provide dependencies via maven artifacts


arnaudgiuliani:
Currently working on it, and reworking some other version variables. Thanks

PR #249, opened 2018-09-28
Possible ConcurrentModificationException fix

owaineevans:
Koin was crashing with concurrency issues - this seems to fix them

### DIFF
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
@@ -82,7 +82,7 @@ class Koin(val koinContext: KoinContext) {
                 registerDefinitions(module())
             }
 
-            logger.info("[modules] loaded ${beanRegistry.definitions.size} definitions")
+            logger.info("[modules] loaded ${beanRegistry.definitionsHashMap.size} definitions")
         }
         logger.debug("[modules] loaded in $duration ms")
         return this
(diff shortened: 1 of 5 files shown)


owaineevans:
Hi, this fix seems to be working for my project and passes all tests.

Sloy:
Hi! Any news on this? 
This bug is stopping us from successfully migrating to Koin :(

arnaudgiuliani:
Actually working also on optimizing search with indexes. Keep in touch.

arnaudgiuliani:
To complete the case, can you provide some test cases please?
(Add it in the `koin-test` project)

arnaudgiuliani:
Do you have a usecase/sample app to reproduce your multi concurrent problem with Koin?

Sloy:
I can try to put something together

Sloy:
I can manage to reproduce it with a test :(
I guess Android instrumentation tests are doing some more nasty stuff for us 😞

owaineevans:
Hi,

I'll have a go at making a test this weekend. But I'm not 100% I can
reproduce it every time. We are using a ThreadPool for our network access
and injecting a single logger class at startup. The crash seemed to only
happen 1% of the time and only in release mode.

When I saw the concurrent modification crash I knew instantly we were
accessing some array from multiple threads at the same time, hence the
patch.

I've been using the fix for a few weeks and the crash has not come back.

So my guess is that to reproduce it you would need to spawn many threads of
the same class that call by inject() at the same time.

I'll see if I've got time this weekend.

Sloy:
There's already a test in `koin-test` for that scenario. So I'm surprised the crash is actually happening.
Maybe it happens in a slightly different scenario?

arnaudgiuliani:
@Sloy That's the point :/

arnaudgiuliani:
I close it. Let's check it with `1.1.x`

ravibhojwani86:
I am at  koin  1.0.2 and see this issue. As per comment above it  is supposed to be fixed in version 1.1.x, but only version  post 1.0.2 I see is 2.0.0, any idea?

PR #252, opened 2018-10-01
Including scopeId param in experimental builders

fredy-mederos:
Including name param in scope functions
Including Tests to validate experimental builders an named scopes


arnaudgiuliani:
Thanks @fredy-mederos !

PR #253, opened 2018-10-01
Including a params-aware experimental "create" function

fredy-mederos:
Including some tests to validate the function when creating instances


arnaudgiuliani:
My decision here is to keep people writing definition if they handle parameters in it. 

Thanks for your contribution.

PR #256, opened 2018-10-02
Made Logger optional.

kuFEAR:
For performance reason it is better to do the Logger as optional. Because if we use the EmptyLogger, It formats the duration message and visualizing the dependency graph even without printing them.


arnaudgiuliani:
Hello @kuFEAR, interesting proposal 👍

paolop:
I'd go for a lazy approach to avoid to evaluate the expressions even if logging is disabled instead of using the safe call operator. IMHO the approach is more "clean" from the invoker perspective. You can find a gist [here](https://gist.github.com/paolop/0bd59e49b33d18d6089fb1bf5488e212) for an example of lazy logging.

arnaudgiuliani:
Integrated into `1.0.2` - PR not merged due to huge rework on main classes.

Thank for contributing

PR #258, opened 2018-10-03
Fixes #246 Optimize BeanRegistry search functions.

### LINKED ISSUE #246: BeanRegistry Search Optimizations
While evaluating Koin for our project we noticed that get / inject times seemed to scale linearly based upon how many dependencies had been supplied to the module(s). For smaller numbers of dependencies (1-100) this is negligible, but it starts becoming more pronounced as your count increases. This is particularly true on older android devices.

I took a look at the underlying implementation and found that the BeanRegistry class maintains a list of definitions and will search through them using Kotlin's `filter` function. This seems to be the primary source of the linear scaling.

I believe if we introduce a couple of maps to the BeanRegistry we can get search complexity to constant(ish) time. I created a quick and dirty proof of concept here: 
https://github.com/ncipollo/koin/pull/2

In this approach I added two maps:
- One for simple class lookups.
- Another for name + class lookups. 

Here are the results of testing with this solution:
- All unit tests pass.
- Search times remain constant regardless of module size. Testing at 1 million dependencies  look ups were typically below a ms, whereas they are 50+ms without the maps.
- This seems to have a negligible impact on module setup time (koin start time was the same before and after change).

**Target Koin project**
koin-core

If this seems like a reasonable approach I would be more than happy to create a legit pull request.

ncipollo:
Fix for #246

### DIFF
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/bean/BeanRegistry.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/bean/BeanRegistry.kt
@@ -16,6 +16,7 @@
 package org.koin.core.bean
 
 import org.koin.core.Koin
+import org.koin.core.instance.DefinitionFilter
 import org.koin.core.name
 import org.koin.core.scope.Scope
 import org.koin.core.scope.isVisibleToScope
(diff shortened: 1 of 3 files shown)


arnaudgiuliani:
Integrated into `1.0.2` - PR not merged due to huge rework on main classes.
Thank for contributing

Issue #322, opened 2018-12-31
koin default implementation without name reference


EslamHussein:
I tried to create 2 Retrofit Clients one of the defaults and the second one has a different implementation like (base URL, interceptors etc... )

I need to inject a default client without using name reference

first client:-

    single<Retrofit> {
    Retrofit.Builder()
        .baseUrl(RemoteConstants.BASE_URL)
        .addCallAdapterFactory(RxJava2CallAdapterFactory.create())
        .addConverterFactory(APIResponseConverter())
        .addConverterFactory(GsonConverterFactory.create(get()))
        .client(get())
        .build()
    }

Second Client:

    single<Retrofit>("retrofit_second") {
        Retrofit.Builder()
                .baseUrl("diffrent url")
                .addCallAdapterFactory(RxJava2CallAdapterFactory.create())
                .addConverterFactory(GsonConverterFactory.create(get()))
                .client(get("SecondOkHttpClient"))
                .build()
    }

My way to inject

    val myModule = module {
    factory { get<Retrofit>().create(FirstAPI::class.java) } // Why Koin did not figure it without providing its default name !?
    factory { get<Retrofit>("retrofit_second").create(SecondAPI::class.java) }
    factory<IMyRemoteDataSource> { MyRemoteDataSource(get(), get()) }
    factory<IMyRepository> { MyRepository(get()) }
    factory { MyUseCase(get()) }
  }

the result is :

    Multiple definitions found for type 'class retrofit2.Retrofit' - Koin can't choose between :
    Single [name='retrofit_second',class='retrofit2.Retrofit']
    Single [name='Retrofit',class='retrofit2.Retrofit']

Why Koin did not get the default Retrofit instance without providing its default name (Retrofit ) !?

    factory { get<Retrofit>().create(FirstAPI::class.java) }

arnaudgiuliani:
What version?

arnaudgiuliani:
to fix it, use the name attribute. You won't be able to resolve it by type as you have several definitions of the same type.

write it: `factory { get<Retrofit>(name = "retrofit_second").create(FirstAPI::class.java) }` or with `retrofit`

EslamHussein:
Koin creates name reference by class name. in this case, why should I give it key or name should koin get it by default.
Also, If I am working on a Big project and we decided to inject another instance of the same class I will do a lot of rework. and actually, this is my case. :)

arnaudgiuliani:
[...]

PR #323, opened 2019-01-01
default injection without name reference

EslamHussein:
default injection without name reference Ask
#322 
https://stackoverflow.com/questions/53988902/koin-default-implementation-without-name-reference


arnaudgiuliani:
I close it. Please check if it has value in Koin 2.0

PR #341, opened 2019-01-28
Add in Koin as an Argument for Android Library and View Model Library

matt-civic:
Solves the problem of developing an Android SDK with its own Koin context
If a local koin is required it can be passed in. 
The default uses the global context to maintain compatibility

### DIFF
--- a/koin-projects/koin-android-viewmodel/src/main/java/org/koin/android/viewmodel/ViewModelResolution.kt
+++ b/koin-projects/koin-android-viewmodel/src/main/java/org/koin/android/viewmodel/ViewModelResolution.kt
@@ -3,12 +3,16 @@ package org.koin.android.viewmodel
 import android.arch.lifecycle.*
 import android.support.v4.app.Fragment
 import android.support.v4.app.FragmentActivity
+import org.koin.core.Koin
 import org.koin.core.KoinComponent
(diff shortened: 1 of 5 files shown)


arnaudgiuliani:
This is not a good approach I think. You can override your koin instance anduse yours, by creating a new implementation of KoinComponent:

```
class SDKKoinComponent : KoinComponent {
    override fun getKoin(): Koin {
        // give your instnce here
    }
}
```

matt-civic:
@arnaudgiuliani  Ok granted that solution works for `inject` and `get` but it doesn't work for `viewModel`. Am I missing something here because I don't see how you can get this method to use the `SDKKoinComponent` you propose creating?

```kotlin
private fun <T : ViewModel> makeViewModelProvider(
    vmStore: ViewModelStore,
    parameters: ViewModelParameters<T>
): ViewModelProvider {
    return ViewModelProvider(
        vmStore,
        object : ViewModelProvider.Factory, KoinComponent {
            override fun <T : ViewModel> create(modelClass: Class<T>): T {
                return getKoin().get(parameters.clazz, parameters.name, null, parameters.parameters)
            }
        })
}
```

arnaudgiuliani:
@pietrini check the beta-3 source, fix has been included.

PR #342, opened 2019-01-28
Fix contributors link

Jawnnypoo:
Right now it 404s


Jawnnypoo:
Looks like this is a dupe of https://github.com/InsertKoinIO/koin/pull/307. You all should probably merge it so this doesn't keep happening 😄

Issue #354, opened 2019-02-08
stopKoin doesn't clear single/scoped instances (Koin 2.0)


erickok:
**Describe the bug**
When stopping Koin, singleton instances are not cleared (released). Starting again will hence return the old instances. We consider this a bug, since we explicitly stop the Koin GlobalContext.

**To Reproduce**
```
fun main(args: Array<String>) {
    // Start first time
    startKoin { modules(modules) }
    val thing = GlobalContext.get().koin.get<Thing>()
    println("Have a $thing")

    // Stop and start again
    stopKoin()
    startKoin { modules(modules) }
    val newThing = GlobalContext.get().koin.get<Thing>()
    println("Have a $newThing")
}

val modules = module {
    single { Thing() }
}

class Thing
```
This prints:
```
Have a Thing@30dae81
Have a Thing@30dae81
```

The cause is that singleton values are stored inside their `SingleInstance`, which are never released since these are bound to their (module) definitions, as long as you have a static definition (i.e. `val myModules = module {}` which is the advertised/recommend way e.g. https://beta.insert-koin.io/ ). In other words, since my definition is not renewed, the singleton value is always kept.

One 'fix' is to create new definitions every invocation of `startKoin { modules(...) }`.
```
val modules get() = module {
    single { Thing() }
}
```
Nnotice the `get()` here. This prints as expected:
```
Have a Thing@30dae81
Have a Thing@1b2c6ec2
```

**Expected behavior**
Stopping the global context (via `stopKoin()`) should release all instances of my singletons.

**Koin project used and used version (please complete the following information):**
koin-core version 2.0.0 beta 1

arnaudgiuliani:
sure 👍

arnaudgiuliani:
I published a fix while you worked on it also. Check the current code.

PR #355, opened 2019-02-08
Stopping Koin should release all instances, cf #354

erickok:
Upon clearing the bean registry (when stopping Koin), also release all kept definition instances.


arnaudgiuliani:
Hello @erickok 

I have done something for it in a previous commit, pushed last week. In the current design I don't want to use the `release` of an instance for single instances, as we don't release singletons. We just make the final clear stuff.

I propose you to check the actual code and double check if I forgot something. 

I close it, but we can continue to discuss about it. Thanks

erickok:
Sure that works and maes sense. I created a separate PR #371 which adds the unit test I wrote to check the fix and for any future regression.

PR #365, opened 2019-02-15
Use parameter name to retrieve same ViewModel from ViewModelProvider

pietrini:
Let users have differents sharedViewModel with same class name. Plus it's more viewModel friendly to use a key instead of rely on a single shared factory instance.


arnaudgiuliani:
@pietrini check the beta-3 source, fix has been included.

Issue #366, opened 2019-02-15
Recreated scope provides the same instance for scoped component


yarkh:
I want to get new scoped instances after scope recreation, even for the same scope id.
Sample code:
```
scope(SCOPE_NAME) {
    scoped { ComponentA() }
}
```
```
val scope1 = getKoin().createScope(scopeId, SCOPE_NAME)
val comp1 = scope.get<ComponentA>()
scope1.close()

val scope2 = getKoin().createScope(scopeId, SCOPE_NAME)
val comp2 = scope.get<ComponentA>()
```

Expected comp1 != comp2 as scope2 was created from scratch, but it is the same instance.

koin-core version 2.0.0-beta-1

erickok:
Think this is a duplicate of #354

yarkh:
Looks like bug in ScopeDefinition release method
```
internal fun release(instance: ScopeInstance) {
    definitions.filter { it is ScopedInstance<*> }.forEach { it.instance.release(InstanceContext(scope = instance)) }
}
```
Should be:
```
internal fun release(instance: ScopeInstance) {
    definitions.filter { it.instance is ScopedInstance<*> }.forEach { it.instance.release(InstanceContext(scope = instance)) }
}
```

feresr:
Hi, I'm still having this issue:


```
abstract class BaseActivity<S : State> : AppCompatActivity() {

    abstract val viewModel: BaseViewModel<S>
    protected val activityScope = getKoin().getOrCreateScope(this::class.getFullName())

    abstract fun onStateChanged(state: S)

    final override fun onStart() {
        super.onStart()
        viewModel.register(::onStateChanged)
        onStartInternal()
    }

    final override fun onStop() {
        onStopInternal()
        viewModel.unregister()
        super.onStop()
    }

    override fun onDestroy() {
        if (!isChangingConfigurations) {
            viewModel.onCleared()
            activityScope.close()
        }
        super.onDestroy()
    }

    protected open fun onStartInternal() {}
    protected open fun onStopInternal() {}
}
```


```
class MainActivity : BaseActivity<HomeState>() {

    override val viewModel: MainViewModel by activityScope.inject { parametersOf(this) }

```

Launching this activity again will provide the same `MainViewModel` instance

PR #368, opened 2019-02-16
Fix ScopeDefinition release

yarkh:
Fix for #366


koin-projects/koin-core/src/main/kotlin/org/koin/core/scope/ScopeDefinition.kt
```diff
@@ -46,7 +46,8 @@ data class ScopeDefinition(val scopeName: String, val module: Module) {
     }
 
     internal fun release(instance: ScopeInstance) {
-        definitions.filter { it is ScopedInstance<*> }.forEach { it.instance.release(InstanceContext(scope = instance)) }
+        definitions.mapNotNull { it.instance as? ScopedInstance<*> }
```
arnaudgiuliani: why not filter operator here?

yarkh: Saved an extra instance call. But I agree, filter would be more readable here.

PR #370, opened 2019-02-17
Fixes #275 Add AutoCloseKoinTestRule.

### LINKED ISSUE #275: Auto-close test rule
In our project we tend to prefer composing behavior into our tests via test rules rather than inheriting behavior from parent classes. The primary benefit of this is that we can mix and match behavior without being constrained to a specific class hierarchy. 

It would be great if Koin provided a test rule with similar functionality to AutoCloseKoinTest. I've created a proof of concept here: https://github.com/ncipollo/koin/pull/3/files.

**Target Koin project**
koin-test

If this seems like it's worth including in the project I can create a legit PR based on the proof of concept linked above.

ncipollo:
Fix for #275.


arnaudgiuliani:
I was expecting a rule to help start/stop koin container, not replace extensions injection there.

ncipollo:
Hey @arnaudgiuliani,

I'm not sure that I understand your feedback. Are you saying that I should update the rule to also support starting Koin? Also, should it continue to use the global context or is there a better way to manage that in 2.0.0?

The rule does stop koin container (stop is called after each test). My rational for having it implement KoinTest is so that if a user choses to use a rule based approach in their tests they will still have access to `inject`, etc.

erickok:
Having the rule also start Koin would be great, passing it a KoinApplication maybe?

ncipollo:
Sure, I'll take a look at having the start Koin as well.

ncipollo:
👍 Will now also start koin. I included a couple of static factory methods. One creates via a `KoinApplication` and the other uses `KoinAppDeclaration`.

arnaudgiuliani:
thanks, let me check that

arnaudgiuliani:
seems fine 👍 

good job

ncipollo:
@arnaudgiuliani Thanks for merging! 

Just a quick note - I had targeted this PR against the 2.0.0 branch. Not sure if that's still the active branch for 2.0.2 or if these changes should be cherry picked down to master.

arnaudgiuliani:
it's ok for that. I will have to make new branches in few days

ncipollo:
Hey @arnaudgiuliani, would you like me to open a PR into 2.0.2? It looks like that's the latest patch release branch right now.

arnaudgiuliani:
@ncipollo yep, go for it 👍

Issue #374, opened 2019-02-20
startKoin logger dsl overrides level


erickok:
When startin Koin with a custom logger, the level gets overridden.
```kotlin
startKoin {
  logger(logger = PrintLogger(Level.ERROR))
}
assertEquals(KoinApplication.logger.level, Level.ERROR)
```
```
java.lang.AssertionError: 
Expected :INFO
Actual   :ERROR
```
While easily fixable, this feels very unnatural to me and I suspect would suprise most users.

I suggest to remove the log level parameter, to remove any doubt, make the API simple and any desired logger level requires a non-empty explicit logger anyway (PrintLogger, AndroidLogger, custom, slf4j, while the defualt shoudl IMHO be EmptyLogger).

(`koin-core version 2.0.0 master branch`)

arnaudgiuliani:
continue this with your PR

PR #375, opened 2019-02-20
Don't override logging level due to logger default parameter, cf #374

erickok:
Instead, an explicit printLogger() function was added for easy setting of printer logger with explicit level.

### DIFF
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
@@ -83,15 +83,18 @@ class KoinApplication private constructor() {
 
     /**
      * Set Koin Logger
-     * @param level
      * @param logger - logger
      */
-    @JvmOverloads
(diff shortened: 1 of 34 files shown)


koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
```diff
@@ -83,16 +83,19 @@ class KoinApplication private constructor() {
...
         KoinApplication.logger = logger
-        KoinApplication.logger.level = level
         return this
     }
 
+    /**
+     * Set Koin to use [PrintLogger], by default at [Level.INFO]
+     */
+    @JvmOverloads
+    fun printLogger(level: Level = Level.INFO) = this.logger(PrintLogger(level))
```
erickok: This is the primary DSL change

arnaudgiuliani: Ok I see the problem with my first proposed DSL value

PR #379, opened 2019-02-22
Mockless and deep checkModules

### LINKED ISSUE #304: KoinTest.checkModules skips module definitions with parameters
**Describe the bug**
If a module definition contains a parameter it will not be verified by `KoinTest.checkModules` (in most cases). Here's what seems to be happening:
1. Depending upon how the parameter is accessed, a `NoParameterFoundException`, a `NoSuchElementException`, or a `IndexOutOfBoundsException` is thrown.
2. This exception is caught and logged in `SandboxInstanceHolder.create`.
3. We jump up the call stack before verifying the actual module definition.

Here is a unit test which demonstrates the issue:
https://github.com/ncipollo/koin/blob/97f5c74d32187e4f264e002951e756dcad5843ed/koin-projects/koin-test/src/test/kotlin/org/koin/test/module/CheckModulesTest.kt#L143

**To Reproduce**
Run the unit test found here: https://github.com/ncipollo/koin/blob/97f5c74d32187e4f264e002951e756dcad5843ed/koin-projects/koin-test/src/test/kotlin/org/koin/test/module/CheckModulesTest.kt#L143

**Expected behavior**
The parameters list ideally won't throw an exception in this case and we can properly validate the module definition. We could maybe mock the parameter list and have it return a mock for any type requested?

erickok:
cf and fixes #304

Reworks `checkModules` function to test real definitions and instantiatability of them. Instead of sandboxing, skipping parameterized dependencies and mocking instances, we try to instantiate the real dependencies as declared, supporting parameter-bound declarations through parameter creators.

See compact but complete example in ModuleCheckTest.


koin-projects/examples/android-mvp/src/test/java/fr/ekito/myweatherapp/ModuleCheckTest.kt
```diff
@@ -1,48 +1,59 @@
...
 import org.mockito.Mockito.mock
+import kotlin.reflect.KClass
 
 /**
  * Test Koin modules
  */
 class ModuleCheckTest : KoinTest {
 
-    val mockedApplication = mock(Application::class.java)
+    private val mocks = module(override = true) {
```
erickok: We only have to mock the (platform API-bound) `JsonReader` now. Also, the real properties and parameters are used to instantiate all declared dependencies.

koin-projects/koin-test/src/test/kotlin/org/koin/test/CheckModulesTest.kt
```diff
@@ -60,7 +61,9 @@ class CheckModulesTest {
                         single { (s: String) -> Simple.MyString(s) }
                     }
             )
-        }.checkModules()
+        }.checkModules(mapOf(
+                Simple.MyString::class to { parametersOf("param") }
```
erickok: Allows checking of parameter-bound dependencies wiht stubbed/test arguments.

koin-projects/examples/android-mvp/src/main/kotlin/fr/ekito/myweatherapp/di/app_module.kt
```diff
@@ -29,7 +29,7 @@ val weatherAppModule = module {
     factory<WeatherListContract.Presenter> { WeatherListPresenter(get(), get()) }
 
     // Presenter for Detail View
-    scope<DetailActivity> {
```
arnaudgiuliani: Why change the scope here? Presenter is tied to the DetailActivity's scope.

koin-projects/examples/android-mvp/src/test/java/fr/ekito/myweatherapp/ModuleCheckTest.kt
```diff
@@ -1,48 +1,59 @@
...
+        single<JsonReader> { mock(JsonReader::class.java) }
+    }
+    private val mockProperties = mapOf(
+            Properties.SERVER_URL to "http://mock/server/url/"
+    )
+    private val parameterCreators = mapOf<KClass<*>, ParametersCreator>(
+            DetailContract.Presenter::class to { parametersOf("id") }
```
arnaudgiuliani: could be nice to reduce the writing with reified function here, to help create params

koin-projects/examples/android-mvp/src/main/kotlin/fr/ekito/myweatherapp/di/app_module.kt
erickok: Sorry, fixed

koin-projects/examples/android-mvp/src/test/java/fr/ekito/myweatherapp/ModuleCheckTest.kt
erickok: Done, I added a `parameterCreatorsOf` DSL-like function

arnaudgiuliani:
one question @erickok, I initially launch this way of making checking by mocks to allow people to check dependency graph even if the internal object fail. For example, an object based on Android context can explode on construction. Whereas with this sandbox/mock approach, we catch the standard exception and continue to follow the dependency graph links.

erickok:
Yes, that is a difference. I believe that if you mock everything, you are not testing anything, becasue a definition is only as good as its instantiatability. As for unavailability of classes, such as Context in JVM tests against an Android project, you can still mock these. (I did so in the MVP and MVVM example projects, but you removed those.)

In some of our projects we don't even mock Context, but Stub them, that is, we have a `single<Context> { StubContext() }` and StubContext is a `class StubContext(...) : Context()` which requires no mocking framework at all.

erickok:
@arnaudgiuliani Merged again, but 2 tests are failing:
```
    @Test
    fun `check a scoped module and ext deps`() {
        koinApplication {
            printLogger(Level.DEBUG)
            modules(
                    module {
                        single { Simple.ComponentB(get()) }
                        scope(named("scope")) {
                            scoped { Simple.ComponentA() }
                        }
                    }
            )
        }.checkModules()
    }
```
Is this test even supposed to succeed? How can `ComponentB`, which dependends on `ComponentA`, be created if it doesn't get a scope, which is needed to create `ComponentA`?

arnaudgiuliani:
@erickok I've just reworked some optimization to make it blazing fast. We just have a conflict on some tests.

I removed the MVVM example to make it back into a separated repository. I keep some minimal test project in the main one.

erickok:
Sure, 👍. I updated again.

I really do understand that this 'checkModules' is a different approach to testing the DI graph than sandboxing, but in my practical experience the sandboxing approach is too 'light'. It is much too easy to have mistakes in your setup and have the test pass. That's why I propose this alternative. On top of it, it is simply less dependent on mocking and that is always a plus IMHO.

In the end, it is your decision. However, I strongly suggest 2.0 and not 2.1 for release, as it is basically a breaking change in the (testing) API.

arnaudgiuliani:
nice, let me check that 👍

arnaudgiuliani:
Thanks for your job @erickok 

I've also a bit updated it to simplify the writing:
```
checkModules {
            create<Simple.MyString> { parametersOf("param") }
            create<Simple.MyString>(UpperCase) { qualifier -> parametersOf(qualifier.toString()) }
        }
```

Issue #348, opened 2019-02-04
Provide a shutdown hook for koin-ktor


viniciusccarvalho:
Currently there's no hook from ktor to allow `StandaloneContext` to stop. Ktor however provides ApplicationEvents: https://ktor.io/servers/lifecycle.html#monitor-events that could be used to subscribe to stop events and then gracefully stop the `StandaloneContext`

arnaudgiuliani:
@viniciusccarvalho if you have any proposal about better Ktor integration, just go 👍

viniciusccarvalho:
@arnaudgiuliani it will be my pleasure to contribute. I also noted that ktor is on a very old version currently (at least on master) 0.9.2, I can also raise a PR to bump its version and repo locations

arnaudgiuliani:
Let's go with your PR then 👍

PR #397, opened 2019-03-13
Ktor enhancements

### LINKED ISSUE #378: Use standard Ktor Install feature
Ok this may be a bit of a pet peeve of mine, but is there a reason why Koin can't be a Feature for Ktor and then just be installed like any other feature?

The proposal is instead of having something like:

```
install(DefaultHeaders)
    installKoin(listOf(common, repositories, services))
```

Have something like 
```
install(DefaultHeaders)
install(Koin){
koin -> {
modules = listOf(common, repositories, services))
}
}
```

It makes it look more aligned with Ktor API, it's a not a big deal I get it, but with 2.0 coming soon, this seems to be something people may look into it.

I volunteer to even submit a PR with the change if this feature makes sense.

Regard

viniciusccarvalho:
Fixes #378, #348 
Hi all, a couple of things on this PR:

1. Upgraded version of Ktor from 0.9.2 (non GA) -> 1.1.3 (latest GA)
2. Added a Feature as described here: https://ktor.io/advanced/features.html 
* I believe using Ktor native approach for extending it via `install(<feature>)` really should be the way to go, eventually all features get registered on the framework Registry, and using a separate way to bootstrap Koin would never allow such visibility.

3. Added a callback to application events to allow the new lifecycle support of beans to be invoked when users stops the application

This is a very **raw** PR, looking for some feedback, in order to be deemed mergeable I still would like to : 

1. Deprecate (or remove) `installKoin`
2. Update the docs
3. Add some testing :)

Please let me know your initial thoughts

### DIFF
--- a/koin-projects/gradle/versions.gradle
+++ b/koin-projects/gradle/versions.gradle
@@ -13,7 +13,7 @@ ext {
     bintray_version = "1.8.4"
 
     // Ktor
-    ktor_version = '0.9.2'
+    ktor_version = '1.1.3'
 
     // Test
     junit_version = "4.12"
--- a/koin-projects/koin-ktor/build.gradle
+++ b/koin-projects/koin-ktor/build.gradle
@@ -15,6 +15,7 @@ dependencies {
 
     // Ktor
     compile "io.ktor:ktor-server-core:$ktor_version"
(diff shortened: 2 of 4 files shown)


arnaudgiuliani:
Thanks!

PR #406, opened 2019-03-22
Callbacks before closing scope and with scope as parameter

Pitel:
Maybe you still want to do something with the scope in those callbacks.


Pitel:
Idea: Maybe pass the scope as parameter to the callback?

arnaudgiuliani:
yeah, add it to the callback

PR #481, opened 2019-06-03
KoinApplication modules method vararg support

TSurkis:
Converted the KoinApplication `modules(modules: Module): KoinApplication` method to a more convenient `modules(vararg modules: Module): KoinApplication` method.
According to the official Kotlin vararg documentation: https://kotlinlang.org/docs/reference/functions.html#variable-number-of-arguments-varargs

In addition to providing an instance of a list of modules, users could also type their modules separated by a comma for a more convenient way.

list way:
`modules(listOf(moduleOne, moduleTwo, moduleThree))`

vararg way:
`modules(moduleOne, moduleTwo, moduleThree)`

single module way:
`modules(moduleOne)`

### DIFF
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
@@ -41,7 +41,7 @@ class KoinApplication private constructor() {
      * @param modules
      */
     fun modules(modules: Module): KoinApplication {
-        return listOf(modules)
+        return modules(listOf(modules))
     }
     
     /**
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
@@ -35,7 +35,15 @@ class KoinApplication private constructor() {
     internal fun loadDefaults() {
         koin.scopeRegistry.loadDefaultScopes(koin)
     }
-
+    
+    /**
+     * Load definitions from a single module
+     * @param modules
+     */
+    fun modules(modules: Module): KoinApplication {
+        return listOf(module)
+    }
+    
     /**
      * Load definitions from modules
      * @param modules
@@ -45,7 +53,7 @@ class KoinApplication private constructor() {
     }
 
     /**
-     * Load definitions from modules
+     * Load definitions from a list of modules
      * @param modules
      */
     fun modules(modules: List<Module>): KoinApplication {
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
@@ -41,7 +41,7 @@ class KoinApplication private constructor() {
      * @param modules
      */
     fun modules(modules: Module): KoinApplication {
-        return listOf(module)
+        return listOf(modules)
     }
     
     /**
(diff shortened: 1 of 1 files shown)


koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
```diff
@@ -40,8 +40,8 @@ class KoinApplication private constructor() {
      * Load definitions from modules
      * @param modules
      */
-    fun modules(modules: Module): KoinApplication {
```
arnaudgiuliani: I would like to keep the one module version, vararg & list if possible

TSurkis: Completed 👍

koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
```diff
@@ -35,19 +35,27 @@ class KoinApplication private constructor() {
...
         koin.scopeRegistry.loadDefaultScopes(koin)
     }
-
+    
     /**
-     * Load definitions from modules
+     * Load definitions from a single module
      * @param modules
      */
     fun modules(modules: Module): KoinApplication {
```
cortinico: You should probably update the name of the parameter here:
```suggestion
    fun modules(module: Module): KoinApplication {
```

arnaudgiuliani: yes

arnaudgiuliani: please add this

TSurkis: This was already completed in one of the previous [commits](https://github.com/InsertKoinIO/koin/pull/481/commits/ca701ebffbc572b5e62383588087faf5c1077bad)

TSurkis:
> @TSurkis Maybe it makes sense also to re-add the method `loadKoinModules` and `unloadKoinModules` from GlobalContext that had a similar behavior?
> 
> See #487 for more context on this

I rather keep this PR small and up to its name, which is just adding vararg support back.
For your proposal, we need a different PR

koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
cortinico: > This was already completed in one of the previous commits

I'm not sure I got your answer @TSurkis. My comment was about renaming the parameter to `module` (singular) as it's only one in this method overload.

cortinico:
> I rather keep this PR small and up to its name, which is just adding vararg support back.

I generally disagree here. I think it would be confusing for library users to have the following scenario: 4 methods with `vararg modules` were removed in `2.0.1` (breaking change). Only of them is reverted in the following version.

From my point of view, either we revert all of them, or we provide a real explanation why only one got reverted rather than all of them. Happy to hear your feedback on this.

arnaudgiuliani:
I clearly agree with @cortinico as we have to be consistent globally on the API. I prefer to see varargs usage with the high-level API entry also, not only one function.

Kerooker:
I vouch for this feature. It was surprising to me that it didn't exist and I had to do a list. It makes the code a little bit more verbose.

I wasn't here when the performance discussion was made, but perhaps we can have it again?

arnaudgiuliani:
I take care of this directly on `2.0.2` 👍

PR #488, opened 2019-06-06
Add support for overriding on-the-fly declarations

### LINKED ISSUE #477: Override on the fly declarations (declare)

I try to `declare` a single on the fly. However, I might do that multiple times if the latest value only is to be available to Koin.

When I do the following, I get a  `org.koin.core.error.DefinitionOverrideException`, logically, because the definition already exists.

```kotlin
koin.declare(x)

// Some point later in time, x is updated and we want to replace its value

koin.declare(x) // Exception!
```

I want to override any existing declaration, but there's no such option. This seems easy to implement, because `declare` under the hood creates a `BeanDefinition`, which could also use the `override` option. However, I cannot enable override through `declare`.

My current alternative is a singleton:

```kotlin
object Holder {
    lateinit var x: Any
}
```

which I use as a Koin `single { Holder.x }` and every time I update `x` I update `Holder.x`.

**Target Koin project**

Koin 2.0 in an Android lib.

cortinico:
This commit adds supports for passing an optional parameter of type
`Option` in the declare methods, used to declare instances on the fly.

This Option can be used to set the `override` flag to true and allow
subsequent declarations of the same type to succeed.

Fixes #477

### DIFF
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
@@ -115,16 +115,15 @@ class Koin {
      * @param instance The instance you're declaring.
      * @param qualifier Qualifier for this declaration
      * @param secondaryTypes List of secondary bound types
-     * @param options Allows to set [Options] for the [BeanDefinition] that will be created, e.g. you can use it to
-     * override a previous declaration of the same type.
+     * @param override Allows to override a previous declaration of the same type (default to false).
      */
     inline fun <reified T> declare(
             instance: T,
             qualifier: Qualifier? = null,
             secondaryTypes: List<KClass<*>>? = null,
-            options: Options? = null
+            override: Boolean = false
     ) {
-        rootScope.declare(instance, qualifier, secondaryTypes, options)
+        rootScope.declare(instance, qualifier, secondaryTypes, override)
     }
 
     /**
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
@@ -16,7 +16,6 @@
 package org.koin.core
 
 import org.koin.core.KoinApplication.Companion.logger
-import org.koin.core.definition.Options
 import org.koin.core.logger.Level
 import org.koin.core.parameter.ParametersDefinition
 import org.koin.core.qualifier.Qualifier
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
@@ -16,6 +16,7 @@
 package org.koin.core
 
 import org.koin.core.KoinApplication.Companion.logger
+import org.koin.core.definition.Options
 import org.koin.core.logger.Level
 import org.koin.core.parameter.ParametersDefinition
 import org.koin.core.qualifier.Qualifier
@@ -111,15 +112,19 @@ class Koin {
      * Declare a component definition from the given instance
      * This result of declaring a single definition of type T, returning the given instance
      *
-     * @param instance
-     * @param qualifier
+     * @param instance The instance you're declaring.
+     * @param qualifier Qualifier for this declaration
+     * @param secondaryTypes List of secondary bound types
+     * @param options Allows to set [Options] for the [BeanDefinition] that will be created, e.g. you can use it to
+     * override a previous declaration of the same type.
      */
     inline fun <reified T> declare(
             instance: T,
             qualifier: Qualifier? = null,
-            secondaryTypes: List<KClass<*>>? = null
+            secondaryTypes: List<KClass<*>>? = null,
+            options: Options? = null
     ) {
-        rootScope.declare(instance, qualifier, secondaryTypes)
+        rootScope.declare(instance, qualifier, secondaryTypes, options)
     }
 
     /**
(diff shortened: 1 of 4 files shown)


koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
```diff
@@ -111,15 +112,19 @@ class Koin {
...
+     * @param secondaryTypes List of secondary bound types
+     * @param options Allows to set [Options] for the [BeanDefinition] that will be created, e.g. you can use it to
+     * override a previous declaration of the same type.
      */
     inline fun <reified T> declare(
             instance: T,
             qualifier: Qualifier? = null,
-            secondaryTypes: List<KClass<*>>? = null
+            secondaryTypes: List<KClass<*>>? = null,
+            options: Options? = null
```
erikhuizinga: I think it's easier for the user if these `Options` were written like here: https://github.com/InsertKoinIO/koin/blob/12e06086e557cf5d33e5e5e3c7bc52c797515bf1/koin-projects/koin-core/src/main/kotlin/org/koin/dsl/Module.kt#L29

Just booleans (or whatever type), so the users have a clear and consistent API.

cortinico: I would love to know @arnaudgiuliani 's point of view. I can easily change it to have just a:

```kotlin
inline fun <reified T> declare(
            instance: T,
            qualifier: Qualifier? = null,
            secondaryTypes: List<KClass<*>>? = null,
            override: Boolean = false
```

But I thought that having `Options` as a parameter offers an easier way to extend this API in the future.  I'm fine either way.

cortinico:
Yup, fixed

koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
```diff
@@ -16,6 +16,7 @@
 package org.koin.core
 
 import org.koin.core.KoinApplication.Companion.logger
+import org.koin.core.definition.Options
```
arnaudgiuliani: why this import?

koin-projects/koin-core/src/main/kotlin/org/koin/core/scope/Scope.kt
```diff
@@ -19,6 +19,7 @@ import org.koin.core.Koin
 import org.koin.core.KoinApplication
 import org.koin.core.definition.BeanDefinition
 import org.koin.core.definition.DefinitionFactory
+import org.koin.core.definition.Options
```
arnaudgiuliani: why this import?

koin-projects/koin-core/src/main/kotlin/org/koin/core/registry/BeanRegistry.kt
```diff
@@ -123,7 +123,11 @@ class BeanRegistry {
         val secondaryTypeDefinitions: ArrayList<BeanDefinition<*>> = definitionsSecondaryTypes[type]
                 ?: createSecondaryType(type)
 
-        secondaryTypeDefinitions.add(definition)
+        if (definition in secondaryTypeDefinitions){
+            secondaryTypeDefinitions[secondaryTypeDefinitions.indexOf(definition)] = definition
```
arnaudgiuliani: not necessary to add the definition if already in secondary types?

cortinico: If the definition is already in the secondaryTypeDefinitions, we want to replace it.
To understand why this is needed, have a look at this test:

```
@Test
    fun `can declare and override a single with secondary type on the fly`() {

        val koin = koinApplication {
            printLogger()
            modules(emptyList())
        }.koin

        val a = Simple.Component1()
        val b = Simple.Component1()

        koin.declare(a, secondaryTypes = listOf(Simple.ComponentInterface1::class))
        koin.declare(b, secondaryTypes = listOf(Simple.ComponentInterface1::class), override = true)

        assertEquals(b, koin.get<Simple.Component1>())
        assertEquals(b, koin.get<Simple.ComponentInterface1>())
    }
```

The alternative would be to convert `secondaryTypeDefinitions` to be a `Map` rather than an `ArrayList` so definitions can't be added twice.

arnaudgiuliani: yeah, or change it for a Set instead of a map

cortinico: > yeah, or change it for a Set instead of a map

Also with `Set` we would need to do something like:
```
if (definition in secondaryTypeDefinitions){
     secondaryTypeDefinitions.remove(definition)
}
secondaryTypeDefinitions.add(definition)
```
I don't know if that is better 🤔

arnaudgiuliani: with a set, we would replace it directly with `add()`, no?

cortinico: Nope. The `add()` function will not replace if the collection already contains the item we're trying to add. From the MutableCollections.kt KDoc:

```kotlin
    /**
     * Adds the specified element to the collection.
     *
     * @return `true` if the element has been added, `false` if the collection does not support duplicates
     * and the element is already contained in the collection.
     */
    public fun add(element: E): Boolean

arnaudgiuliani: ok :) `definitionsSecondaryTypes` has to stay as a map. Let's keep to your proposal

Issue #467, opened 2019-05-26
How to get viewModel from Koin with SavedStateHandle (new Google announced at I/O 2019)


mihovu:
I'm concerning how to get ViewModel from Koin with SavedStateHandle. 
Here is my DI class:  
**ViewModelModule.kt**
```kotlin
viewModel { ProfileViewModel(userApi = get()) }
```
Now, I want to try using SavedState module for viewModule that is announced from Google I/O 2019.   
Link: https://developer.android.com/topic/libraries/architecture/viewmodel-savedstate

So I declared in my ViewModel as below:  
```kotlin
class ProfileViewModel(
    private val userApi: UserApi, 
    private val stateHandle: SavedStateHandle
) : ViewModel() {
```

Base on Koin concept, I just need define viewModel in ProfileActivity.kt class:  
```kotlin
private val viewModel: ProfileViewModel by inject()
```

But, what should I do with my DI class ViewModelModule.kt to get SavedStateHandle ?

IVIanuu:
I'm pretty sure that this isn't possible at the moment.

mehonhuong:
@IVIanuu 
Thank for the reply.
Basically I think so

lucassales2:
On dagger that's is solved using `@AssistedInject`, can we expect something similar coming to Koin?

arnaudgiuliani:
have to check how to use this `SavedStateVMFactory` with constructor injection

BulatMukhutdinov:
Any updates on this?

arnaudgiuliani:
This feature is open to contribution if you want to propose something 👍

kedzie:
I think we would need to add DSL which passes a savedStateParameter to the bean definition.  I could look at it..

```
viewModelWithState { (savedState: SavedStateHandle, someArgument : String) -> MyViewModel(savedState, someArgument, get(), get()) }
```

kedzie:
I added a PR for this  #496 would appreciate feedback.  I know it forces the lifecycle dependencies to update, maybe we can discuss it.

kedzie:
@arnaudgiuliani I think this will be essential functionality.  is there a bleeding edge branch you use to point to the latest versions of androidx?  If so that would be a better target.  I would also love to here your thoughts on my implementation.  #496

arnaudgiuliani:
ok, will check that

arnaudgiuliani:
closing this, to go on the PR directly

PR #496, opened 2019-06-14
#467 add viewmodel saved state support

kedzie:
This is for issue #467

### DIFF
--- a/koin-projects/koin-androidx-viewmodel/src/main/java/org/koin/androidx/viewmodel/ViewModelResolution.kt
+++ b/koin-projects/koin-androidx-viewmodel/src/main/java/org/koin/androidx/viewmodel/ViewModelResolution.kt
@@ -67,7 +67,7 @@ fun <T : ViewModel> Scope.createViewModelProvider(
 ): ViewModelProvider {
     return ViewModelProvider(
             vmStore,
-            if(beanRegistry.findDefinition(parameters.qualifier, parameters.clazz)!!.isStateViewModel()) {
+            if(beanRegistry.findDefinition(parameters.qualifier, parameters.clazz)?.isStateViewModel() == true) {
                 object : AbstractSavedStateVMFactory(parameters.owner as SavedStateRegistryOwner, parameters.defaultArguments()) {
                     override fun <T : ViewModel?> create(key: String, modelClass: Class<T>, handle: SavedStateHandle): T {
                         return get(parameters.clazz, parameters.qualifier) {
--- a/koin-projects/koin-androidx-viewmodel/src/main/java/org/koin/androidx/viewmodel/ViewModelResolution.kt
+++ b/koin-projects/koin-androidx-viewmodel/src/main/java/org/koin/androidx/viewmodel/ViewModelResolution.kt
@@ -68,7 +68,7 @@ fun <T : ViewModel> Scope.createViewModelProvider(
     return ViewModelProvider(
             vmStore,
             if(beanRegistry.findDefinition(parameters.qualifier, parameters.clazz)!!.isStateViewModel()) {
-                object : AbstractSavedStateVMFactory(vmStore as SavedStateRegistryOwner, parameters.defaultArguments) {
+                object : AbstractSavedStateVMFactory(vmStore as SavedStateRegistryOwner, parameters.defaultArguments()) {
                     override fun <T : ViewModel?> create(key: String, modelClass: Class<T>, handle: SavedStateHandle): T {
                         return get(parameters.clazz, parameters.qualifier) {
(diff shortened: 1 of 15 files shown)


kedzie:
@arnaudgiuliani any thoughts?  I was wondering about the `sharedViewModel`, if we need variations of the methods for those as well.  `sharedViewModelWithState`, I was trying to avoid that.

arnaudgiuliani:
Hello @kedzie , good job.

Could you propose something without adding a new injection function? 

We could pass a param from the `viewModel` dsl keyword like `viewModel(useState = true) { (state : ...) -> YourVIewModel(sate) }` or even propose a new keyword `stateViewModel { ... }`

Don't know what could be the best.

And from the resolution step, we can detect if we have the flag property to use the new factory

kedzie:
Will do !  thanks for taking a look

kedzie:
@arnaudgiuliani Great advice! made the PR much much cleaner.

coro101:
@kedzie I hope this pull request will be merged soon! Great work!

ustadenis:
@kedzie Could you please say whan it will be merged? Thank you.

arnaudgiuliani:
@kedzie one last question, why do we need to expose `defaultArguments: Bundle? = null` in `FragmentExt.kt` or other viewModel() extension?

kedzie:
The [AbstractSavedStateVMFactory](https://developer.android.com/reference/androidx/lifecycle/AbstractSavedStateVMFactory.html) takes a `Bundle defaultArgs` for initial values in the SavedStateHandle.  The optional argument allows those to be passed in.

arnaudgiuliani:
A lazy approach can be interesting effectively. The thing is to avoid conflict with the actual parameters lambda. Could you reuse it also? 🤔 

I mean, reuse the `parametersOf` lambda stuff to pass your bundle.

kedzie:
i can think of 2 ways to pass in the default bundle in the definition parameters:

1. pass Bundle as an additional **first** parameter of definitions.. so something like `by currentScope.viewModel(this ) { parametersOf(Bundle().apply { putString("bling" "blang") }, "vm2") }` . problem is that bundle is optional so when resolving this must detect if first parameter is a bundle?  what if 2nd parameter is also a bundle?  resolving can be complicated..
2. pass Bundle as an additional **last** parameter of definitions.. so something like `by currentScope.viewModel(this ) { parametersOf("vm2", Bundle().apply { putString("bling" "blang") }) }` . if we know the # of parameters expected would could detect if an extra one has been passed in, which is the Bundle.  
3. Could have bundle be passed into the parameters method and populated... but this doesn't really match the rest of koin.. `by currentScope.viewModel(this) { defaultArgs -> 
defaultArgs.putString("bling", "blang")
parametersOf("vm2") }`

_An argument for keeping it the way it is:_

`ViewModelParameters` already has another lambda for ` val from: ViewModelStoreOwnerDefinition? = null`
where `typealias ViewModelStoreOwnerDefinition = () -> ViewModelStoreOwner`

So the defaultArguments bundle as it's own lambda follows that same pattern.

arnaudgiuliani:
@kedzie yeah, not so clean to have this bundle here :/ But then we need those args at the creation of the VM. Perhaps best is to keep it as an optional parameter finally?! 🤔

kedzie:
keep it as a lazy optional parameter I think is best.

kedzie:
I could add some espresso instrumentation tests to run through the `androidx-samples` . so we know at build time all this stuff resolves..

jgavazzisp:
Hi team, any update in when this branch will get merged in?

Thanks

koin-projects/koin-androidx-viewmodel/src/main/java/org/koin/androidx/viewmodel/ViewModelResolution.kt
```diff
@@ -61,9 +67,18 @@ fun <T : ViewModel> Scope.createViewModelProvider(
 ): ViewModelProvider {
     return ViewModelProvider(
             vmStore,
-            object : ViewModelProvider.Factory {
-                override fun <T : ViewModel> create(modelClass: Class<T>): T {
-                    return get(parameters.clazz, parameters.qualifier, parameters.parameters)
+            if(beanRegistry.findDefinition(parameters.qualifier, parameters.clazz)!!.isStateViewModel()) {
```
arnaudgiuliani: avoid !! here please 👍

arnaudgiuliani:
Seems to be fine. Just one comment, to avoid using `!!` operator. @kedzie no need of expresso tests, too heavy to pass.  Good job 👍

PR #535, opened 2019-07-23
Add 'Testing a Koin application with KotlinTest' article

Kerooker:
We recently added an integration between Koin and KotlinTest.

This article was written to help users do that integration


arnaudgiuliani:
You can put it on the top of the list rather ;)

Kerooker:
Nice! Done :)

Issue #550, opened 2019-08-14
Lazy injection crashes from the class starts Koin


budius:
**Describe the bug**
Lazy injections `by inject()` crash if created before starting Koin

**To Reproduce**
Check the [attached minimum reproducible example](https://github.com/InsertKoinIO/koin/files/3500493/KoinBugReport.zip).  

tl;dr (snippet below shouldn't crash, but it does)
```
fun main() {
   val crashy = CrashyInjection(context)
   startKoin {  ... etc }
   Log.d("TAG", "Injected `${crashy.component1}`")
}

class CrashyInjection(context: Context) : KoinComponent {
    val component1: Component1 by inject()
}
```

**Expected behavior**
If the injection is lazy, it should be really lazy and not try to access the Koin component from its declaration

**Koin project used and used version (please complete the following information):**
Used `org.koin:koin-android:2.0.1`, but AFAIK it's on `koin-core`

budius:
See pull request with fixes here https://github.com/InsertKoinIO/koin/pull/551

arnaudgiuliani:
resolved in PR 👍

PR #551, opened 2019-08-14
Make 'by inject()' function very lazy,

budius:
This removes access to getKoin() during 'by inject()' declaration,
and moves all its declaration to a 'lazy{ }' block.
This allows for classes that declare by 'inject()' to be initialised before 'startKoin{}'

Added test for the condition.

Fixes gh-550

### DIFF
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinComponent.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinComponent.kt
@@ -52,7 +52,7 @@ inline fun <reified T> KoinComponent.inject(
         qualifier: Qualifier? = null,
         noinline parameters: ParametersDefinition? = null
 ): Lazy<T> =
-        getKoin().inject(qualifier, parameters)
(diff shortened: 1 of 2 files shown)


budius:
See bug report https://github.com/InsertKoinIO/koin/issues/550 for detailed explanations and minimum reproducible example.

arnaudgiuliani:
Hi @budius 

good catch. Could you check also if any other `inject` function is in the same state? I could have miss it elsewhere 👍

arnaudgiuliani:
thanks 👍

budius:
@arnaudgiuliani 

I've checked on the main areas that my code uses. 

- LifecycleOwnerExt.kt (`by viewModel`) have all properly wrapped in `lazy { }`
- KoinComponent.kt is just this one from the MergeRequest
- Koin.kt redirects Lazy calls to rootScope. But rootScope is `val`, so it's fine.
- Scope.kt have all properly wrapped in `lazy { }`

arnaudgiuliani:
Great 👍

PR #559, opened 2019-08-26
#480 Add viewModel functions for Java compatibility

### LINKED ISSUE #480: ViewModel injection from Java class.
I'm not able to properly retrieve a ViewModel from a Java class. I'm following this guide https://insert-koin.io/docs/2.0/getting-started/android-java/ I suspect this approach does not work for ViewModels since I'm receiving a new instance every time thus loosing the LiveData payload after the Activity gets recreated after a configuration change. Is there somewhere a documentation for injecting ViewModels in Java classes?

This is how the module is declared: 
val unlockScreenViewModelModule: Module = module {
   viewModel { UnlockScreenViewModel<DomainObject>(androidApplication(), get()) } // get() asks Koin to find the dependency among the declared modules (See Koin.kt)
   single {
      @Suppress("UNCHECKED_CAST")
      BusinessRepositoryImpl(get()) as BusinessRepository<DomainObject>
   }
}

jeziellago:
This PR fix #480 
I add some classes for create Java compatibility with `viewModel` and `sharedViewModel` extensions.
Also, I created on sample one activity and fragment in Java, showing the view model injections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,7 @@ class MyApplication : Application() {
 
 ### Articles & resouces about Koin
 
+* [Testing a Koin application with KotlinTest](https://dev.to/kerooker/testing-koin-applications-with-kotlintest-1iip)
 * [Ready for Koin 2.0](https://medium.com/koin-developers/ready-for-koin-2-0-2722ab59cac3)
 * [Migration from Dagger2 to Koin](https://proandroiddev.com/migrating-from-dagger2-to-koin-3b2b3f5285e9)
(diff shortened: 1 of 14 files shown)


arnaudgiuliani:
very good job. I'll merge it soonly 👍 

Did you updated the docs also? Else I will do :)

roelvandun:
Great to see this! When will 2.1.0 be released? Or is it possible to have early access?

arnaudgiuliani:
currently early access: `2.1.0-alpha-1`

PR #568, opened 2019-09-03
Allow enums to be used as Qualifiers (along with String or a class)

Jerbell:
This is a simple qualifier I am using and should be useful to others.  It allows an enum to be used as a qualifier.  This avoid defining equal strings in multiple areas and allows refactoring without any hidden surprises.

### DIFF
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/qualifier/EnumQualifier.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/qualifier/EnumQualifier.kt
@@ -4,8 +4,8 @@ package org.koin.core.qualifier
  * Class which allows enums to be used as Qualifiers.
  * More strict than strings and are better for refactoring.
  * @author Julian Bell
- * @sample get(named(Direction.NORTH))
- * @sample inject(named(Direction.SOUTH))
+ * @sample get(qualifier(Direction.NORTH))
+ * @sample inject(qualifier(Direction.SOUTH))
  */
 data class EnumQualifier<E : Enum<E>>(val value: E) : Qualifier {
     override fun toString(): String {
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/qualifier/EnumQualifier.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/qualifier/EnumQualifier.kt
@@ -1,0 +1,14 @@
+package org.koin.core.qualifier
+
+/**
+ * Class which allows enums to be used as Qualifiers.
+ * More strict than strings and are better for refactoring.
+ * @author Julian Bell
+ * @sample get(named(Direction.NORTH))
+ * @sample inject(named(Direction.SOUTH))
+ */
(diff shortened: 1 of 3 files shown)


erickok:
The idea to have a `Qualifier` interface in the first place was indeed to have (refactor) safety instead of relying on string. Why not use objects instead of an enum? Potentially together with a sealed class if you prefer.

Jerbell:
I'd prefer to use enums rather than Objects.  An ObjectQualifier would work as well.

erickok:
May I ask why you prefer `get(named(Way.Left))` over `get(Way.left)`?

```kotlin
sealed class Way: Qualifier {
  object Left: Way()
  object Right: Way()
}
```

Jerbell:
You do realise the documentation only mentions String qualifiers?
https://insert-koin.io/docs/2.0/quick-references/modules-definitions/#qualifiers
You are asking why I prefer using `get(named(Way.Left))` over `get(Way.Left)` - that's not my point - I prefer defining an `enum class Way { Left, Right }` for it's simplicity, rather than a sealed class which extends Qualifier.
If there are people out there who like & use enums like myself then they will find this useful.

erickok:
I did not realize that and consider that a weakness in the docs then. If I find time I might be able to contribute to better docs (imho better).

I suggest you should rename the `named` function as it is not a named qualifier at all, it is an enum-based qualifier. Perhaps into `qualify(MY_ENUM_VALUE)`.

Jerbell:
Yes - `named` should be renamed.  Actually - I can use `enum` if that's clear?  I thought it would be a reserved word but apparently not in this case.
`private val rsaBits: Int = get(enum(RSA_BITS))`
What do you think?

erickok:
I like `qualify(X)` or `qualified(X)` or `enumQualifier(X)` better, but it is up to the project maintainers.

erickok:
Still, it effectively uses the string name, which gives unexpected non-safe behaviour. Trying to do this throws an exception with `Already existing definition`:
```
factory(enum(Way.LEFT)){ ...}
factory(named("LEFT")){ ...}
```
Worse, trying to get a qualified instance using named, e.g. `get(named("LEFT"))`, will work even if you defined only `factory(enum(Way.LEFT)){ ...}`.

Jerbell:
So I've changed the method to `qualified` - I think it reads the best.  I've also fixed and written more tests around the potential problems.  There's a Koin/module test - hopefully I'm doing that correctly.

arnaudgiuliani:
Hi,

I'm thinking if we could propose like `qualifier()` more than `named()`, which would accept kind of types. We can make an alias and deprecate the current API. 

then `qualifier<Enum>`, `qualifier<KClass>` or `qualifier("a string")` is more consistent for me.

tynn:
Why don't you just define the enum class as a qualifier? It should be sufficient for most use-cases.

```
enum class Way : Qualifier {
    Left, Right
}
```

Jerbell:
I didn't think of having an enum extend an interface.
It seems to work perfectly - thanks!  I'd recommend updating the documentation on Qualifiers.

PR #569, opened 2019-09-03
Added an article about Koin

yolave:
I wrote an article after replace Dagger with Koin. I hope this could be added :)


arnaudgiuliani:
just moved the link to the top 👍

Issue #462, opened 2019-05-21
Is there any way to refer the qualifier inside the definition? 


RodrigoAmaroEng:
Hey guys, i was wondering if is there a way to refer the qualifier to avoid code duplication on my module.

Let's consider the following construction:

```
module {
   single<ClassA>(QualifierA1) { ClassA(get<ClassB>(QualifierB1)) }
   single<ClassA>(QualifierA2) { ClassA(get<ClassB>(QualifierB2)) }
   single<ClassB>(QualifierB1) { ClassB(1) }
   single<ClassB>(QualifierB2) { ClassB(2) }
}
...
get<ClassA>(QualifierA1)
get<ClassA>(QualifierA2)
```
I consider that this construction have some "issues":

1. I'm obligated to create a different qualifier for the same purpose (QualifierA1 and QualifierB1). This is only necessary because when i use a qualifier, it binds to the type meaning i can't use the same qualifier to another type
2. I need to duplicate the ClassA construction even it being the same for both qualifiers just because its dependencies have different qualifiers

In my opinion it would be better if i could write this:

```
module {
   single<ClassA>(proxyQualifier = true) { ClassA(get<ClassB>(qualifier)) }
   single<ClassB>(Qualifier1) { ClassB(1) }
   single<ClassB>(Qualifier2) { ClassB(2) }
}
...
get<ClassA>(QualifierA1)
get<ClassA>(QualifierA2)
```
This way, even thought ClassA injector does not define a qualifier, i could pass it to its dependencies and avoid both code duplication and unnecessary qualifiers. Besides the fact that this example only skips one line of code, i have bigger dependency trees on my project where it would really decrease the amount of code needed

In this example `proxyQualifier` is just a property name to identify that this scope can receive a qualifier to use it on its dependencies.

So i don't know if it's already possible but, if not, would be a great feature to have

erickok:
Can't you use two scoped instances instead of two singletons?

arnaudgiuliani:
also using `bind` operator?

RodrigoAmaroEng:
@erickok, @arnaudgiuliani 
In this example i used `single` just to elaborate the problem. 

I confess i got a little confused on how it can be more "readable" and offer a less verbose solution, but i'm pretty sure i just have to experiment with these proposed solution to clarify on this. After doing i will post my impressions here

RodrigoAmaroEng:
So i've read the documentation about `scope` and `bind` features.

[...]

PR #571, opened 2019-09-03
provide a combination of name (string) and type qualifiers together.

HKhademian:
now you can use qualifiers like:

    factory(named<A>("hi")){ ...}
    factory(named<B>("hi")){ ...}
    factory(named<C>("hi")){ ...}

all together.

### DIFF
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/qualifier/Qualifier.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/qualifier/Qualifier.kt
@@ -14,3 +14,8 @@ fun named(name: String) = StringQualifier(name)
  * Give a Type based qualifier
  */
 inline fun <reified T> named() = TypeQualifier(T::class)
+
+/**
+ * Give a String + Type qualifier
+ */
+inline fun <reified T> named(name: String) = StringTypeQualifier(T::class, name)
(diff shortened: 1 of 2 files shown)


erickok:
Can you give a real-world example? Also, please read #568.

HKhademian:
I implement clean arch. and repository pattern in android apps.

I have `...DataSource`s and for each I have api and db data-sources.

I setup koin to provide singles like:

```kotlin
val moduleA = module {
    single(named("api")) { ... } bind PostDataSource::class
    single(named("db")) { ... } bind PostDataSource::class

    single(named("api")) { ... } bind TagDataSource::class
    single(named("db")) { ... } bind TagDataSource::class
}
```

As you can see, with default qualifiers we cannot define modules like this.

HKhademian:
After I use the patch in my app, it's hard to use `named<A>("hi")` and `named("hi")` and `named<A>()` together in a same kotlin file.

So I suggest either convert `TypeQualifier` to accept names (with null default, maintain code backward compatibility), or change this new Qualifier helper's function name to something like `typeNamed<A>("hi")`

I think if we follow the first approach will be better, because we don't have extra useless code and data structs, ...

erickok:
Indeed this sounds more like a workaround to the issue that qualifiers cannot be used multiple times (for different types).

HKhademian:
yes, I think `StringTypeQualifier` or `TypeQualifier` with string value is a simple solution for #462.

arnaudgiuliani:
Your problem is solved with new resolution engine, as you can have the same qualifier with 2 different type.

Please check with `2.1.0-alpha-6`

Issue #566, opened 2019-08-29
Scope loses old object instances.


haykmelqonyan:
hi @arnaudgiuliani  thanks for good product, found some issue in scopes.

after creating new `id` on same `scope` all old instances in that scope are lost
here is clear explanation.
definition of my scope in modules.

```
scope(named("my_scope")) {
		scoped {
			MyClass()
		}
	}
```
**To Reproduce**

usage in my Activity

```
                 val koin = getKoin()
		val scope1 = koin.createScope("A", named("my_scope"))
		val scope2 =koin.createScope("B", named("my_scope"))

		//get instnace from id A
		val instanceA = scope1.get<MyClass>()
		//get instnace from id B
		val instanceB = scope2.get<MyClass>()
		//get instnace from id A
		val instanceACopy1 = scope1.get<MyClass>()
		
		//instances must be equals. they both are from id A
		println("instanceA === instanceACopy1 ${instanceA === instanceACopy1}")
                //output "instanceA === instanceACopy1 true"
```
this case works correct , output `instanceA === instanceACopy1  true`

but.
lets try this 
```

		val koin = getKoin()
		val scope1 = koin.createScope("A", named("my_scope"))
		val scope2 =koin.createScope("B", named("my_scope"))

		//get instnace from id A
		val instanceA = scope1.get<MyClass>()
		//get instnace from id B
		val instanceB = scope2.get<MyClass>()
		//get instnace from id A
		val instanceACopy1 = scope1.get<MyClass>()

		//what happens here, every time when I create new Id in exact  scope, all old IDs in that scope are removed.
		val scope3 =koin.createScope("C", named("my_scope"))


		val instanceACopy2 = koin.getOrCreateScope("A", named("my_scope")).get<MyClass>()

		//instances must be equals. they both are from id A
		println("instanceA === instanceACopy1 ${instanceA === instanceACopy1}")//output "instanceA === instanceACopy1 true"

		//instances must be equals , the both are from id A, but they are not.
		println("{instanceA === instanceACopy2 ${instanceA === instanceACopy2}") 
                //output "instanceA === instanceACopy2 false"
```

so problem is ,every time when I create new `id` in exact `scope` all previosly created objects (ScopeDefinitions) in that `id` are removed

**Expected behavior**
I need to receive same object instances from same `id` every time befor I close it.

**Koin project used and used version (please complete the following information):**
```
 org.koin:koin-core:2.0.1
org.koin:koin-android:2.0.1
[...]

Issue #492, opened 2019-06-08
Crash during Fragment replace for scoped ViewModel


egorikftp:
**Describe the bug**
I was found an issue when I try to open fragment and then _replace_ with the same fragment.
In our prod application, it's a popular case. 

`    java.lang.IllegalStateException: Definition without any InstanceContext - [type:Factory,scope:'com.egoriku.testapplication.fragment.CrashFragment', primary_type:'com.egoriku.testapplication.viewmodel.TestViewModel']
        at org.koin.core.definition.BeanDefinition.resolveInstance(BeanDefinition.kt:71)
        at org.koin.core.scope.Scope.resolveInstance(Scope.kt:165)
        at org.koin.core.scope.Scope.get(Scope.kt:128)
        at org.koin.androidx.viewmodel.ViewModelResolutionKt$createViewModelProvider$1.create(ViewModelResolution.kt:66)
        at androidx.lifecycle.ViewModelProvider.get(ViewModelProvider.java:135)
        at androidx.lifecycle.ViewModelProvider.get(ViewModelProvider.java:103)
        at org.koin.androidx.viewmodel.ViewModelResolutionKt.getInstance(ViewModelResolution.kt:43)
        at org.koin.androidx.viewmodel.ViewModelScopeResolutionKt.getViewModel(ViewModelScopeResolution.kt:14)
        at org.koin.androidx.viewmodel.ext.android.ScopeExtKt.getViewModel(ScopeExt.kt:86)
        at com.egoriku.testapplication.fragment.CrashFragment$$special$$inlined$viewModel$1.invoke(ScopeExt.kt:96)
        at com.egoriku.testapplication.fragment.CrashFragment$$special$$inlined$viewModel$1.invoke(ScopeExt.kt)
        at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
        at com.egoriku.testapplication.fragment.CrashFragment.getModel(CrashFragment.kt)
        at com.egoriku.testapplication.fragment.CrashFragment.onViewCreated(CrashFragment.kt:26)
`

**To Reproduce**
I have prepared a sample demo project with a crash.

https://github.com/egorikftp/KoinBug

Steps to reproduce the behaviour:
1. Install the demo app
2. Click on 'Replace fragment'
3. Then click to button once again

**Expected behaviour**
New fragment instance created without a crash.

**Koin project used and used version:**
`koin version 2.0.1`

**Note**
[...]

Issue #447, opened 2019-05-13
Proper way to inject Activity Scope dependencies in its Child Fragments


epardus:
Is there any recommended way to inject Activity scope dependencies in its child fragments?
Currently, I use something like this:
`val userService by lazy {activity?.currentScope?.get<UserServices>()} ` 

, BUT I think it is not the right way to do so.

arnaudgiuliani:
use inject from scope: `val userService by activity?.currentScope?.inject<UserServices>()`

epardus:
Thanks for your quick answer.
Unfortunately, it is leading to an `NPE` 
we are using `v2.0.0-GA2`
```
Caused by: java.lang.NullPointerException: Attempt to invoke interface method ‘java.lang.Object kotlin.Lazy.getValue()’ on a null object reference
       at de.msahakyan.poc.koinscopes.ui.HeadersFragment.getUser(Unknown Source:25)
       at de.msahakyan.poc.koinscopes.ui.HeadersFragment.onResume(HeadersFragment.kt:32)
       at androidx.fragment.app.Fragment.performResume(Fragment.java:2498)
       at androidx.fragment.app.FragmentManagerImpl.moveToState(FragmentManager.java:1501)
       at androidx.fragment.app.FragmentManagerImpl.moveFragmentToExpectedState(FragmentManager.java:1784)
       at androidx.fragment.app.FragmentManagerImpl.moveToState(FragmentManager.java:1852)
       at androidx.fragment.app.FragmentManagerImpl.dispatchStateChange(FragmentManager.java:3269)
       at androidx.fragment.app.FragmentManagerImpl.dispatchResume(FragmentManager.java:3241)
       at androidx.fragment.app.FragmentController.dispatchResume(FragmentController.java:223)
       at androidx.fragment.app.FragmentActivity.onResumeFragments(FragmentActivity.java:538)
       at androidx.fragment.app.FragmentActivity.onPostResume(FragmentActivity.java:527)
       at androidx.appcompat.app.AppCompatActivity.onPostResume(AppCompatActivity.java:172)
       at android.app.Activity.performResume(Activity.java:7141)
       at android.app.ActivityThread.performResumeActivity(ActivityThread.java:3556)
       at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3621)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2862)
       at android.app.ActivityThread.-wrap11(Unknown Source:0)
[...]

Issue #428, opened 2019-04-17
Cleaner way to inject activities in a multiple activity app?


plusmobileapps:
At my day job, I'm working in a project that spans across multiple activities and there are some classes in the project that depend on an activity such as Google Pay. The team is looking for a cleaner way to inject activities into our dependencies and we were looking at Koin as a replacement to our home grown DI. 

I have done research which landed me on this Koin [issue](https://github.com/InsertKoinIO/koin/issues/49) on Github which has a few suggestions on injecting activities It seemed that the towards the end of the issue, the suggested approach is to use injection parameters like so:

```kotlin
//LocationModule.class
factory { (activity: MainActivity) -> FusedLocationProviderClient(activity) }

//MainActivity.class
val provider: FusedLocationProviderClient by inject { parametersOf(this@MainActivity) }
```

This is great if the dependency is actually being used in the view, however in my use case it is injected to an Interactor further back in the architecture and could be tedious to have to declare anything that requires an activity dependency in two locations instead of one. 

So one approach I took to try to accomplish this was to create a kotlin object that actually keeps track of the current activity with a stack. There were some concerns about the correct activity being supplied and potential memory leaks. 

```kotlin
object ActivityStack {
    private val stack = Stack<BaseActivity>()
    val currentActivity: BaseActivity
        get() = stack.peek()
    fun push(activity: BaseActivity) = stack.push(activity)
    fun pop() = stack.pop()
}

//then declared inside of a module 
factory<GooglePayDataSource> { GooglePayDataSourceImpl(ActivityStack.currentActivity) }
factory { GooglePayInteractor(get()) }

```

The second approach was creating an extension function on our `BaseActivity` to declare any activity dependencies. But this required me to override any dependencies declared in Koin which in my use case don't think would be an issue because the Interactor is also instantiated in the creation of the activity so it has a reference to `GooglePayDataSource` already. Although I could see this being an issue potentially if the dependency is not needed right away. 

```kotlin
fun BaseActivity.declareActivityDependencies() {
    val activityModule = module {
        factory<GooglePayDataSource>(override = true) { GooglePayDataSourceImpl(this@declareActivityDependencies) }
    }

    loadKoinModules(activityModule)
}
```

There were also concerns of the above approach having memory leaks since the [documentation](https://insert-koin.io/docs/2.0/documentation/koin-android/index.html#_taming_the_android_lifecycle) on scoping state the following. 

[...]

PR #592, opened 2019-10-03
Attaching objects to a scope instance

aschattney:
So first of all, thanks to the authors of this library. We use it in one of our Android apps
and it works out pretty good.

This pull request mainly targets two new features:
- Attaching objects to a scope instance during creation.
    Create scopes with arguments #231
    Proper way to inject Activity Scope dependencies in its Child Fragments #447
    Cleaner way to inject activities in a multiple activity app? #428
- Declaring child scopes including hierarchy validation. (Cannot create nested scope #200)

And additionally provides fixes for the following issues:
- ~~Scope loses old object instances. #566~~ See Pull Request: [601](https://github.com/InsertKoinIO/koin/pull/601)
- ~~Crash during Fragment replace for scoped ViewModel. #492~~ See Pull Request: [602](https://github.com/InsertKoinIO/koin/pull/602)

# Attaching objects to a scope instance during creation.
A new method is introduced in the ScopeSet API which creates an object scope:

```kotlin
module {
    objectScope<MainActivity> {
        // instance refers to the object instance of type MainActivity
        scoped { KeyboardService(instance) }
    }
}
```

# Declaring child scopes including hierarchy validation.

The parent scope id must be provided during scope instance creation.
An exception is thrown when the scopeName of the parent scope instance doesn't match 
with the scopeName of the parent scope definition.

```kotlin
module {
    objectScope<MainActivity> {
        scoped { KeyboardService(instance) }
        childObjectScope<MainFragment> {
             scoped { ToolbarController(instance, get<KeyboardService>()) }
             scoped { ScopedFragmentService(get<MainActivity>()) }
        }
    }
}
```


erickok:
Could you explain a use case, something that you could not do before with Koin and can do now with object scopes? This is a pretty big PR, with 31 commits and 71 files changed, but from my first impression is that you can do this already with the existing scope API?

```kotlin
module {
    scope(named<MainActivity>()) {
        scoped { (activity: Activity) -> KeyboardService(activity) }
    }
}
```
and then inject it with
```kotlin
private val keyboardService: KeyboardService by currentScope.inject { parametersOf(this) }
```

Now, indeed that requires inject ing the `Activity` as parameter, but that is not a problem for me. In fact if I see your modified sample project you inject a [`Controller`](https://github.com/InsertKoinIO/koin/blob/0ec0fe12c182a360a7cb710cf00b484c989058fb/koin-projects/examples/android-samples/src/main/java/org/koin/sample/android/components/scope/Controller.kt) into the [`MVVMFragment`](https://github.com/InsertKoinIO/koin/blob/0ec0fe12c182a360a7cb710cf00b484c989058fb/koin-projects/examples/android-samples/src/main/java/org/koin/sample/android/mvvm/MVVMFragment.kt#L26) but you can just supply the activity as parameter no?

aschattney:
Thank you for your comment. I added an example in the androidx-samples module which illustrates the problem.

erickok:
As I suspected you do all this because you want to inject the activity somewhere deep down the dependency tree, yet since you don't control its instantiation you have to pass it all the way down with parameters. I agree this is ugly, but would argue that the activity requirement deep in your hiearchy is code smell. What is `SomeService` going to do with the activity.

Again, not saying its wrong or not clean, just making sure we introduce a feature not for one specific annoyance or niche use case.

arnaudgiuliani:
>Again, not saying its wrong or not clean, just making sure we introduce a feature not for one specific annoyance or niche use case.

Yes, the same feeling here 👍

aschattney:
> [...], but would argue that the activity requirement deep in your hiearchy is code smell. [...]
I agree with you here.

We favor components which use the activity or fragment instead of the activity/fragment performing the actions directly, so we avoid encoding the logic there. These components are then composed as necessary, so we actually don't inject the activity directly in some fragment scoped component, but a component which uses the activity.
This approach provides us more flexibility during integration tests (which is really nice with Koin btw), where we can setup the app as necessary.

Suppose we have a feature where we scan a qr code. With something like a PermissionService we can control the current permission state and how the system answers to permission requests, without actually interacting with the permission system during the integration test.

When developing a kotlin mpp app, you basically need abstractions like this, if you want such logic as permission handling in your shared code module.

arnaudgiuliani:
Hey @aschattney 

I would like here to propose to split the fixes and your feature proposal. The hard part here is that you've already done lots of work, and this is a massive one.

No problem to prototype things. After that we need to agree on how we want to integrate that globally in the project.

One of the difficulties about Scope is that for now, the definition in the DSL is easily confused with runtime behavior.

Perhaps we need to step back about Scopes, and think about granularity and level: for now scope is only one level.

Don't know what do you think about that @erickok ?

erickok:
I see how it can be neccesary for a component like `PermissionService` (or the implementation of such an interface at least) to have access to the `Activity`. Still, coupling this to the scoping seems convenient rather than obvious. What about having such as `PermissionService` accept an `ActivityProvider` instead, which would be a component that always has a reference to the current activity? (I would implement this component by having itself register as `Application.ActivityLifecycleCallbacks`.)

erickok:
As for the sub-scopes. Could you give an example of a case where the current scoping feature is insufficient? In your example, you scope `ToolbarController` to the fragment but that seems overkill - why not have a factory?

aschattney:
@arnaudgiuliani 
It took quite some time but i don't mind throwing away/changing parts of the implementation.
I will split the implementation into several merge requests to separate the fixes and features.

Actually, i think a special concept of an 'object scope' might not be required since it's actually possible with the current api to declare a bean definition on a scope instance on the fly. The only difference would be that this specific definition is not available after declaring scope but directly after creation of the scope instance which doesn't make a difference in the end.
So actually most of the changes would be regarding the parent/child scope feature.

Regarding the parent/child scope feature:
I think i can get rid of the whole generic 'S: Scope' stuff which is the source of most of the changes. The intention behind this was to be able to directly access the property 'instance' declared in an object scope which actually isn't accessible anymore in a bean definition.

@erickok 

> What about having such as PermissionService accept an ActivityProvider instead, which would be a component that always has a reference to the current activity?
This would disallow accessing the ActivityProvider during `onCreate()` in the activity. The callback function `onActivityPreCreated` has been added for API 29 to the `ActivityLifecycleCallbacks` interface but it's not available in earlier versions.

> Could you give an example of a case where the current scoping feature is insufficient?

In general, it feels like having half of the configuration and wiring defined using koin and half of wiring happening at the injection site.

Let's take [FragmentScenario](https://developer.android.com/training/basics/fragments/testing) as an example. It provides the ability to test a fragment standalone. Unfortunately it's not possible to define the activity class it uses for the fragment under test. So during resolution of the fragments dependencies, it might pass the activity as a parameter to a bean definition. When the parameter type is an interface which the actual Activity implements, then the resolution will fail because the activity which is used in this test does not implement this specific interface. When the fragment just relies on the object graph, then it would be possible to override necessary definitions without the fragment requiring to resolve the parameters itself.

I will think about a more detailed example.

Using Dagger 2, the resolver can automatically resolve dependencies from parent components when necessary. Using the [@BindsInstance](https://dagger.dev/users-guide.html#binding-instances) annotation it's possible to make components like activity or fragment available in the corresponding dagger components object graph. I see more people switching from dagger to koin when similar features are available, with nearly zero setup along with easy usage.

erickok:
Sure I can agree with scoped objects being available in the dependency graph.

arnaudgiuliani:
I close this PR to split the work on several topics. 👍

makaroffandrey:
What is the status of this? I see it is marked "Done", but I do not see any mention of it in the docs. This API also seems to be missing from `2.1.0-beta1`.
Am I missing something?

moizalidv:
@arnaudgiuliani is it possible to send parameters to scoped?

makaroffandrey:
@moizalidv
> is it possible to send parameters to scoped?

As far as I know, there is no official way to do this, but there is a workaround that I have already mentioned [here](https://github.com/InsertKoinIO/koin/issues/428#issuecomment-595950513)
You add this extension function:
```kotlin
inline fun <reified T> Scope.addInstance(instance: T): Scope {
    val definition = BeanDefinition(
        scopeDefinition = this._scopeDefinition,
        primaryType = T::class,
        kind = Kind.Single,
        definition = { instance }
    )
    val factory = SingleInstanceFactory(this._koin, definition)
    (this._instanceRegistry.instances as HashMap<IndexKey, InstanceFactory<*>>)[indexKey(
        T::class, null
    )] = factory
    return this
}
```
Then you configure your scope as if your parameter is already available
```kotlin
scope<ParameterForScope> {
    scoped {
        ClassThatNeedsParameters(parameterForScope = get())
    }
}
```
And then, you call `.addInstance` on your scope before using it.
```kotlin
val parameterForScope = ParameterForScope();
val scope = parameterForScope.getOrCreateScope().addInstance(parameterForScope)
```

PR #594, opened 2019-10-04
Update CoffeeApp.kt

ermolnik:
erroneous double call


arnaudgiuliani:
was for perfs testing, thanks.

PR #596, opened 2019-10-07
Add varargs on load/unload modules

charbgr:
Hi!

I think this addition will make the api more fluent!

Let me know what do you think!

### DIFF
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
@@ -48,6 +48,14 @@ class KoinApplication private constructor() {
      * Load definitions from modules
      * @param modules
      */
+    fun modules(vararg modules: Module): KoinApplication {
+        return modules(modules.toList())
+    }
+
+    /**
+     * Load definitions from modules
+     * @param modules
+     */
     fun modules(modules: List<Module>): KoinApplication {
         if (logger.isAt(Level.INFO)) {
             val duration = measureDurationOnly {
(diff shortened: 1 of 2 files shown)


koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
```diff
@@ -44,6 +44,14 @@ class KoinApplication private constructor() {
         return modules(listOf(modules))
     }
 
+    /**
+     * Load definitions from modules
+     * @param modules
+     */
+    fun modules(vararg modules: Module): KoinApplication {
+        return modules(modules.toList())
```
dwaxemberg: The `modules` function's only use of the list is to call `loadModulesAndScopes` which takes an `Iterable`, so we can change `modules` to also take an `Iterable` and we'll no longer need to worry about converting to a list.

charbgr: so you say to change `module` function to take an `Iterable` ?

dwaxemberg: Yeah. Change `fun modules(modules: List<Module>): KoinApplication {` on line 59 to `fun modules(modules: Iterable<Module>): KoinApplication {`

arnaudgiuliani:
Did you check about `GlobalContext` in 2.1.0? It already has many overload about those functions

charbgr:
> Did you check about `GlobalContext` in 2.1.0? It already has many overload about those functions

I didn't see `2.1.0`. Should I make a PR with `Iterable` ?

arnaudgiuliani:
Why we would need `Iterable` in those API? 🤔  We already have overload for List, Varag ... 

is it to allow any kind of "collection"?

charbgr:
> Why we would need `Iterable` in those API? 🤔 We already have overload for List, Varag ...
> 
> is it to allow any kind of "collection"?

Well I don't have idea why we should allow pass any other kind of collection! We can close the PR since I haven't see the `varargs` on `modules` in 2.1.0.

PR #597, opened 2019-10-08
Adding Qualifier information when failing to find bean definition

dwaxemberg:
Include Qualifier in error message when we fail to find an instance in the bean registry.

This makes debugging much easier when using multiple definitions of the same type, but with different qualifiers.

For example:
```
fun example() {
    val app = koinApplication {
        modules(
                module {
                    single(named("one")) { 1 }
                    single(named("two")) { 2 }
                }
        )
    }
    val koin = app.koin
    val one: Int = koin.get(named("one"))
    val two: Int = koin.get (named("two"))
    val three: Int = koin.get(named("three"))
}
```
Previously, this will throw an error that reads: 
`Exception in thread "main" org.koin.core.error.NoBeanDefFoundException: No definition for 'java.lang.Integer' has been found. Check your module definitions.`

Checking the module definitions we can see that there are in fact Integer definitions, which leads to confusion.

After the changes we have:
`Exception in thread "main" org.koin.core.error.NoBeanDefFoundException: No definition for 'java.lang.Integer' with qualifier 'three' has been found. Check your module definitions.`
Now it's clear that we are missing a definition for "three".

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ jobs:
         
       # run tests!
       - run: 
-          command: ./koin-projects/gradlew -p ./koin-projects clean test -Dorg.gradle.parallel=false
+          command: ./koin-projects/gradlew -p ./koin-projects clean test
 
 workflows:
   version: 2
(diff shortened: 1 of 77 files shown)


erickok:
Nice. Good work.

arnaudgiuliani:
thanks

PR #602, opened 2019-10-10
#492 Avoid closing definitions when a scope instance is closed

aschattney:
Since there might be other open scope instances with the same scope name. Fixes gh-492

### DIFF
--- a/koin-projects/koin-core/src/test/kotlin/org/koin/dsl/BeanLifecycleTest.kt
+++ b/koin-projects/koin-core/src/test/kotlin/org/koin/dsl/BeanLifecycleTest.kt
@@ -1,8 +1,10 @@
 package org.koin.dsl
 
-import org.junit.Assert.assertEquals
+import org.junit.Assert.*
 import org.junit.Test
 import org.koin.Simple
+import org.koin.core.definition.BeanDefinition
+import org.koin.core.error.NoBeanDefFoundException
 import org.koin.core.qualifier.named
 
 class BeanLifecycleTest {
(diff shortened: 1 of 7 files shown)


koin-projects/koin-core/src/test/kotlin/org/koin/dsl/BeanLifecycleTest.kt
```diff
@@ -1,8 +1,10 @@
 package org.koin.dsl
 
-import org.junit.Assert.assertEquals
+import org.junit.Assert.*
```
ppjeziellago: Could you fix this wildcard import?

koin-projects/koin-core/src/test/kotlin/org/koin/dsl/BeanLifecycleTest.kt
```diff
@@ -109,4 +111,82 @@ class BeanLifecycleTest {
...
+        assertNotNull(scopeA2Value)
+        assertNotEquals(scopeA1Value, scopeA2Value)
+
+        scopeA1.close()
+
+        assertEquals(scopeA2Value, scopeA2.get<Simple.Component1>())
+        try {
+            scopeA1.get<Simple.Component1>()
+            fail("Should not be able to resolve the instance from scope A because scope A is closed.")
+        } catch(e: NoBeanDefFoundException) {
```
ppjeziellago: I think you could replace try/catch block, only using `@Test(expected = NoBeanDefFoundException::class)`.

ppjeziellago:
@arnaudgiuliani could you check theses changes? Thanks! :)

koin-projects/koin-core/src/test/kotlin/org/koin/dsl/BeanLifecycleTest.kt
aschattney: This might lead to a false positive because `Simple.Component1` is resolved twice and should only throw the exception at the second call. So the exception is not expected for the first retrieval of `Simple.Component1`.

arnaudgiuliani:
this has to be review with new internal design, about definition/instance structure. It would help have a more clear parallel/locking strategy for Koin.

arnaudgiuliani:
Fixed in current `2.1.0-alpha-6`

PR #603, opened 2019-10-10
Allow overriding definitions in scopes by loading modules

aschattney:
Previously, the definitions were not added when they were already present. Specifying the override parameter didn't work because only primaryType and qualifier are evaluated in the `hashCode()` and `equals()` method in the `BeanDefinition` class.


arnaudgiuliani:
Can you help and check if you have this problem in current master version? (2.1.0-alpha-6)

arnaudgiuliani:
Can you check if this issue is still there on `2.1.0-alpha-7`?

aschattney:
The issue is resolved :thumbsup:

arnaudgiuliani:
good :)

PR #604, opened 2019-10-11
Mockito in Android instrumentate tests and declareMock()

erickok:
Test that Mockito can be used with Android instrumentation tests and koin-test.


arnaudgiuliani:
thanks 👍

PR #605, opened 2019-10-11
Nested scopes

aschattney:
Introduces the mechanism of nested scopes in `koin-core`.


arnaudgiuliani:
Hello @aschattney 

Thank you for your proposal, I clearly appreciate your interest in Koin. 

With such breaking feature, I would recommend exposing and sharing your ideas before going into tons of changes. Then we can agree on DSL, API, Samples & use cases.

From my point of view, Scope are a way to describe a portion of graph with a different lifecycle than the classic single/factory one. But the DSL describe in Koin the logical structure, not the runtime one.

If you are interested in, I can propose that we design the `Scope Links` together. Scope links would be a way to say that a scope depends on others. We would start then with a very simple mechanism before into too much deep things.

What do you think?

aschattney:
@arnaudgiuliani Yeah sure, we can do that

arnaudgiuliani:
Closing this PR for Scope Links instead.

PR #611, opened 2019-10-13
[Docs] Remove ekito.fr article links from README

Haider8:
ekito.fr’s server IP address could not be found


arnaudgiuliani:
fixed in current master

Issue #615, opened 2019-10-20
Remove single function from ScopeSet class.


Raiden18:
Remove single function from ScopeSet class.

**Describe the solution you'd like**
Just remove single function from ScopeSet class. This method is depricated, throws IllegalStateException  and does nothing. Moreover ScopeSet doesn't even implement any interfaces that lead it to implementing the single method. It seems that single function is just useless.

**Target Koin project**
Koin-Core.

arnaudgiuliani:
Hello,

check my answer here: https://github.com/InsertKoinIO/koin/pull/618#issuecomment-544950163

PR #618, opened 2019-10-22
Remove single fun

AminBakhtiyari:
Remove a single function from the ScopeSet class.

for: #615


arnaudgiuliani:
No, sorry. This function is to block any kind of attempt to use `single` in a scope. With Kotlin, and how we have such functions, it's hard to prevent such usage.

Raiden18:
But what about Lint? It's possible to notify user not to use the single function.

PR #623, opened 2019-10-25
Migrates project to AndroidX and Q

rafaeltoledo:


anticafe:
I cannot believe Koin still not support AndroidX yet :D

erickok:
I'm sorry, but while it is reasonable to upgrade the examples to Android,
it is simply wrong to upgrade the explicit non-AndroidX libraries. There
are separate modules for AndroidX. Did you totally miss these?

https://github.com/InsertKoinIO/koin/tree/master/koin-projects

Op za 26 okt. 2019 06:13 schreef anticafe <notifications@github.com>:

arnaudgiuliani:
hum ... we have already AndroidX libraries 🤔

Issue #527, opened 2019-07-12
DeclareMock not working with instrumentation test android


hardikp888:
**Describe the bug**
declareMock<>() in koin not working with instrumentation test-case, Same test case working in Junit.

**To Reproduce**

my gradle file contain

```
implementation "org.koin:koin-android:2.0.1"
testImplementation "org.koin:koin-test:2.0.1"
androidTestImplementation "org.koin:koin-test:2.0.1"
```

My testcase

```
class DemoTest : KoinTest {

    class ComponentA
    class ComponentB(val a: ComponentA)

    @Test
    fun runDemoTest() {
        startKoin {
            modules(
                module {
                    single { ComponentA() }
                    single { ComponentB(get()) }
                })
        }
        declareMock<ComponentA>()

        // retrieve mock
        assertNotNull(get<ComponentA>())

        // is built with mocked ComponentA
        assertNotNull(get<ComponentB>())
    }
}
```

DemoTest is available in both androidTest and test Directory.But testcase in test directory working fine while testcase in androidTest fail with below error.

`org.koin.core.error.InstanceCreationException: Could not create instance for [type:Single,primary_type:'com.hardik.demos.DemoTest$ComponentA'] at org.koin.core.instance.DefinitionInstance.create(DefinitionInstance.kt:61) at org.koin.core.instance.SingleDefinitionInstance.get(SingleDefinitionInstance.kt:40) at org.koin.core.definition.BeanDefinition.resolveInstance(BeanDefinition.kt:70) at org.koin.core.scope.Scope.resolveInstance(Scope.kt:165) at org.koin.core.scope.Scope.get(Scope.kt:128) at com.hardik.demos.DemoTest.runDemoTest(DemoTest.kt:73) at java.lang.reflect.Method.invoke(Native Method)`

**Expected behavior**
DeclareMock should be work with androidTest also.

arnaudgiuliani:
for this problem, I would require some help to investigate 👍 
thanks

erickok:
Mocking on Android (instrumented tests) does not work with vanilla Mockito, as it would try to modify class byte data, yet Android uses dex. Specifically, Koin's koin-test project declares `org.mockito:mockito-inline` as dependency, but you'll need `mockito-android`.

You can override the mockito dependency and use `mockito-android`. After that it will still work as expected. See #604 for example (and actual test that tests this is possible).

Issue #392, opened 2019-03-11
A more Kotlin-friendly mocking framework in Koin JUnit Test


roccadev:
Mockito is not Kotlin-friendly and it's quite a pain using it for testing: see for example the ugly \`when\`, or the Java-ish builder patterns or the lack of `ArgumentMatcher.anyNonNull()`.

Furthermore, it's also a pain trying to put the right dependencies to make it work in instrumented tests for Android.

There are mocking libraries built with Kotlin and *for* Kotlin, like [Mockk](https://mockk.io/), that I'm personally using with absolute delight, that could be better in harmony with the nature of Koin.
Mockk is DSL style and with its extensive usage of infix notation, it's very fluent too. 
It's also featured in [Kotlin Academy](https://blog.kotlin-academy.com/).

I checked out the koin-test code and making the switch is really a matter of a couple of lines.
However, maybe there's some better approach, like having different artifacts that use different dependencies, so that existing code with Mockito is not affected.

erickok:
Mocking might be (from the Koin perspective) unneeded at all; see #379

arnaudgiuliani:
@erickok interesting. And then we would avoid any mock stuff, and let people do on their own?

We would remove `declareMock` with just a `declare` function to let declare everything

erickok:
Oh yes, I think a mocking framework has no place in Koin codebase, IMHO, except perhaps with some helper functions, but those should be in a different module, such that you can use koin-test without getting the dependency. Not everyone wants to use or needs a mocking framework, and some use a different one.

arnaudgiuliani:
sure 👍

arnaudgiuliani:
I just reworked the internals to fix many problem. Here is the new shape of mocking (without any more link to mockito):

```kotlin
class CoffeeMakerTest : KoinTest {

    private val coffeeMaker: CoffeeMaker by inject()
    private val heater: Heater by inject()

    @get:Rule
    val koinRule = KoinTestRule.create {
        printLogger(Level.DEBUG)
        modules(coffeeAppModule)
    }

    @get:Rule
    val mockProvider = MockProviderRule.create { clazz ->
        Mockito.mock(clazz.java)
    }

    @Test
    fun testHeaterIsTurnedOnAndThenOff() {
        declareMock<Heater> {
            given(isHot()).will { true }
        }

        coffeeMaker.brew()

        verify(heater, times(1)).on()
        verify(heater, times(1)).off()
    }
}
```

[...]

PR #651, opened 2019-11-24
Deprecate direct usage of Mockito in koin-test

tynn:
Related to #392 and #527 I rewrote `declareMock` to rely on `declare` to override the dependency.


erickok:
Excellent, I strongly agree with this and with the code in this PR. 👍

arnaudgiuliani:
New internal design is also in progress ... I will explain more about that when it's ready.

tynn:
@arnaudgiuliani due to the deprecation with this version it should be easier to keep it and remove it with the code. It could be set to `compileOnly` though.

arnaudgiuliani:
Just rewrite internals to fix scope, memory/immutable graph problem ... 
And then put the hands in koin-test:

Here is the new shape of mocking (without any more link to mockito):

```kotlin
class CoffeeMakerTest : KoinTest {

    private val coffeeMaker: CoffeeMaker by inject()
    private val heater: Heater by inject()

    @get:Rule
    val koinRule = KoinTestRule.create {
        printLogger(Level.DEBUG)
        modules(coffeeAppModule)
    }

    @get:Rule
    val mockProvider = MockProviderRule.create { clazz ->
        Mockito.mock(clazz.java)
    }

    @Test
    fun testHeaterIsTurnedOnAndThenOff() {
        declareMock<Heater> {
            given(isHot()).will { true }
        }

        coffeeMaker.brew()

        verify(heater, times(1)).on()
        verify(heater, times(1)).off()
    }
}
```

arnaudgiuliani:
@tynn could you check on current master, if it's away from your proposal?

arnaudgiuliani:
as the mockito is externalised, I kindof think we should provide default mock providers:

this has to be done by hand:

```
 @get:Rule
    val mockProvider = MockProviderRule.create { clazz ->
        Mockito.mock(clazz.java)
    }
```

tynn:
I think the `CoffeeMakerTest` is still failing. Somehow `declare` doesn't have any effect. I'll have a look at it again. @arnaudgiuliani maybe you notice the issue directly.

tynn:
I think I found the last issue. All tests are passing. `declare` uses `intance::class` instead of the reified `T::class`. Also I removed all deprecated inline functions. They are not used on runtime and on compiletime the new ones would be prefered anyhow.

arnaudgiuliani:
It has to be part of the documentation to setup the Mock provider now

PR #652, opened 2019-11-25
Remove kotlin-reflect dependency

linroid:
Koin doesn't use it, but it will increase about 1MB size


arnaudgiuliani:
good 👍

deviant-studio:
is it already released? i still see `kotlin-reflect` in the dependencies

PR #666, opened 2019-12-14
gh-577 Koin/Ktor: Configure stopping behaviour

### LINKED ISSUE #577: Koin/Ktor: Configure stopping behaviour
Currently the Ktor Koin feature does this

```kotlin
override fun install(pipeline: Application, configure: (KoinApplication).() -> Unit): Koin {
    val application = KoinApplication.create()
    application.apply(configure)
    GlobalContext.start(application)
    pipeline.environment.monitor.subscribe(ApplicationStopping) {
        GlobalContext.stop()
    }
    return Koin()
}
```
the problem I have run into is that during the ApplicationStopping lifecycle, I may have other components which need to be shutdown, and which are resolved via Koin.

I can work around it by eager loading all of the components, but the root issue is that I want to ensure I can manage all their shutdowns before the GlobalContext is stopped.

Would it be acceptable to have a configurable property that controls this behaviour, eg.

```kotlin
install(Koin) {
      autostop = false   // defaults to true to maintain existing intuitive behaviour
      modules(appModule)
}
```

```kotlin
if (autostop) {
    pipeline.environment.monitor.subscribe(ApplicationStopping) {
        GlobalContext.stop()
    }
}
```

this enables me to do things like
```kotlin
environment.monitor.subscribe(ApplicationStopping) {
    environment.log.info("Stopping services")
    elasticService.disconnect()
    GlobalContext.stop()
}
```

BorzdeG:
* Added Koin application events
* Fixed #577
* Upgrade Ktor version
* Upgrade Gradle version
* refactoring KoinFeature.kt
* Added .editorconfig


BorzdeG:
Changed the main branch from master to 2.1.0

PR #687, opened 2020-01-04
Improve module plus function

mohsenoid:
This new function allows you to plus a list of modules to the current module.
I faced this case in my project and I believe it should be inside the API.


arnaudgiuliani:
thanks @mohsenoid 👍

Issue #658, opened 2019-11-30
Koin module does not return after installation in Ktor application


BorzdeG:
**Describe the bug**

**To Reproduce**
```kotlin
fun Application.main(){
    install(Koin){
        println("1")
    }
    featureOrNull(Koin)?.let{
        println("2")
    }
}
```
console output:
```
1
```

**Expected behavior**
console output:
```
1
2
```

**Koin project used and used version (please complete the following information):**
 [e.g]: `koin-core version 0.9.2`
koin-server version 1.2.6
koin-ktor version 2.1.0-alpha-3

PR #688, opened 2020-01-04
parent koin

NkD:
This is idea for solution of multi module Android app. 
If Android have these modules:
:core
:feature:login
:feature:market

then :core modul can have global koin and every :feature module can have own koin module binded with parent koin from :core. Parent koin contains shared instances (singletons) as OkHttp and Retrofit.

### DIFF
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
@@ -36,7 +36,7 @@ import kotlin.reflect.KClass
  *
  * @author Arnaud Giuliani
  */
-class Koin {
+class Koin(val parent: Koin? = null) {
     val _scopeRegistry = ScopeRegistry(this)
     val _propertyRegistry = PropertyRegistry(this)
     var _logger: Logger = EmptyLogger()
@@ -200,7 +200,9 @@ class Koin {
      * @param qualifier
      */
     fun getOrCreateScope(scopeId: ScopeID, qualifier: Qualifier): Scope {
-        return _scopeRegistry.getScopeOrNull(scopeId) ?: createScope(scopeId, qualifier)
+        return _scopeRegistry.getScopeOrNull(scopeId)
+                ?: parent?._scopeRegistry?.getScopeOrNull(scopeId)
+                ?: createScope(scopeId, qualifier)
     }
 
     /**
@@ -210,7 +212,9 @@ class Koin {
      */
     inline fun <reified T> getOrCreateScope(scopeId: ScopeID): Scope {
         val qualifier = TypeQualifier(T::class)
-        return _scopeRegistry.getScopeOrNull(scopeId) ?: createScope(scopeId, qualifier)
+        return _scopeRegistry.getScopeOrNull(scopeId)
+                ?: parent?._scopeRegistry?.getScopeOrNull(scopeId)
+                ?: createScope(scopeId, qualifier)
     }
 
     /**
@@ -219,6 +223,7 @@ class Koin {
      */
     fun getScope(scopeId: ScopeID): Scope {
         return _scopeRegistry.getScopeOrNull(scopeId)
+                ?: parent?._scopeRegistry?.getScopeOrNull(scopeId)
                 ?: throw ScopeNotCreatedException("No scope found for id '$scopeId'")
     }
 
@@ -228,6 +233,7 @@ class Koin {
      */
     fun getScopeOrNull(scopeId: ScopeID): Scope? {
         return _scopeRegistry.getScopeOrNull(scopeId)
+                ?: parent?._scopeRegistry?.getScopeOrNull(scopeId)
     }
 
     /**
@@ -243,7 +249,9 @@ class Koin {
      * @param defaultValue
      */
     fun <T> getProperty(key: String, defaultValue: T): T {
-        return _propertyRegistry.getProperty<T>(key) ?: defaultValue
+        return _propertyRegistry.getProperty<T>(key)
+                ?: parent?._propertyRegistry?.getProperty<T>(key)
+                ?: defaultValue
     }
 
     /**
@@ -252,6 +260,7 @@ class Koin {
      */
     fun <T> getProperty(key: String): T? {
         return _propertyRegistry.getProperty(key)
+                ?: parent?._propertyRegistry?.getProperty(key)
     }
 
     /**
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinApplication.kt
@@ -27,9 +27,9 @@ import org.koin.core.time.measureDuration
  *
  * @author Arnaud Giuliani
  */
-class KoinApplication private constructor() {
+class KoinApplication private constructor(val parent: Koin? = null) {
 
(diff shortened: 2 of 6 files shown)


NkD:
Please take this as a suggestion.
It's possible that I haven't held any contracts in Koin that I don't even know about. In this case, just get inspired by this pull-request.
Anyway, this is functionality that is missing a lot in Koin and I would need it for my application.

BorzdeG:
Maybe my PR #658  will be more convenient to use? See an example https://github.com/BorzdeG/koin/tree/gh-658/koin-projects/examples/multimodule-ktor

NkD:
@BorzdeG I was looking at it, but I don't really understand the connection with my problem.

BorzdeG:
@NkD We solved one problem differently - to add modules to the configuration.
Only I solved it for the Ktor module.

Maybe in the test "https://github.com/BorzdeG/koin/blob/gh-658/koin-projects/koin-ktor/src/test/kotlin/org/koin/ktor/ext/KoinFeatureTest.kt#L109" will be more obvious?

NkD:
@BorzdeG I am not familliar with Ktor, but if I understanding completelly, it's looks like something else then my solution, because You add modules to same Koin instance. 
In my solution I have parent Koin instance and child Koin instance. Child Koin have link to parent Koin and I can get singleton from parent Koin if I call get() on child Koin. Child try find singleton and if not found then delegate to parent. This is very important mechanism in multi module environment (AFAIK).

arnaudgiuliani:
Hello @NkD 

thanks for your submission. Having Koin modules spread around Gradle modules is already something you can do: https://doc.insert-koin.io/#/koin-core/modules?id=using-several-modules

Why do you need to have separated Koin instance for your modules then? 🤔

NkD:
@arnaudgiuliani because I need singletons which are shared accross whole Android application. 
For example OkHttp, Retrofit, OauthRepository, etc. This singletons are in "parent" Koin, and every android module (feature) have own Koin instance with own repositories and viewmodels etc, and this Koin have link to parent which is used as fallback.

NkD:
@arnaudgiuliani You can see simplest example (parent/koin) in https://github.com/InsertKoinIO/koin/pull/688/commits/fff5405fb175b4eb750b1b873574ab07323dbea9#diff-217cb11f473108237340eda0b8178fa2R13

parent koin will be in :core android module
and koin will be in some :feature module

arnaudgiuliani:
I understand your examples, but why do you need separated Koin instances?

BorzdeG:
Can’t you solve the parent-child problem using the scope API?

NkD:
@arnaudgiuliani because I can't write something like this:
`@Test
    fun notWorking() {
        val parentModule = module {
            single { Simple.ComponentA() }
        }
        val child1 = koinApplication {
            modules(
                    listOf(
                            parentModule,
                            module {
                                single { Simple.ComponentB(a = get()) }
                            }
                    )
            )
        }.koin
        val child2 = koinApplication {
            modules(
                    listOf(
                            parentModule,
                            module {
                                single { Simple.ComponentB(a = get()) }
                            }
                    )
            )
        }.koin
        val child1_A = child1.get<Simple.ComponentA>()
        val child2_A = child2.get<Simple.ComponentA>()
        val child1_B = child1.get<Simple.ComponentB>()
        val child2_B = child1.get<Simple.ComponentB>()
        Assert.assertEquals(child1_A, child2_A)
        Assert.assertEquals(child1_A, child2_B.a)
        Assert.assertEquals(child2_A, child1_B.a)
        Assert.assertNotEquals(child1_B, child2_B)
    }`

NkD:
@arnaudgiuliani this "notWorking" test can be solution for my problem, but it not work, because parentModule not work in child2 and I don't know why. Maybe because it was used and internal structures can't be reused? And if worked, get I same instance of ComponentA in child1 and child2 ? I don't think so.

NkD:
@BorzdeG no, it's different problem

BorzdeG:
> no, it's different problem

Yes, I noticed it because once deleted my comment...

NkD:
Koin DSL is problematic in thinking about. Using terminology which override global terms like applcation and module. In Android world are this terms used for something else. And if I talk about what I need I use this terms from both worlds and it's not clear explain exactly what I mean.

Android module != Koin module
Android application != KoinApplication
etc...

NkD:
.. but unit tests explains exactly what I need.
Simple say: I need shared singletons across the app and separate instances in each Android module, but they can use shared singletons

NkD:
If there is any machanism in Koin that I overlook and can be used in this way, then please direct me.
I didn't find anything like that. Maybe _linkedScope in 2.1.0, but I don't fully understand it here and it definitely doesn't have an external access API.

arnaudgiuliani:
With Scope Links, you could share some instance: 

```kotlin
class ComponentA
class ComponentB(val a: ComponentA)
```

```kotlin
 @Test
    fun `shared linked scope `() {
        val koin: Koin = startKoin {
            modules(
                    module {
                        scope<A> {
                            scoped { Simple.ComponentB(get()) }
                        }
                        scope<B> {
                            scoped { Simple.ComponentB(get()) }
                        }
                        scope<C> {
                            scoped { Simple.ComponentA() }
                        }
                    })
        }.koin


        val scopeA = koin.createScope<A>()
        val scopeB = koin.createScope<B>()
        val scopeC = koin.createScope<C>()
        scopeA.linkTo(scopeC)
        scopeB.linkTo(scopeC)

        val compb_scopeA = scopeA.get<Simple.ComponentB>()
        val compb_scopeB = scopeB.get<Simple.ComponentB>()

        assertNotEquals(compb_scopeA, compb_scopeB)
        //shared ComponentA instance
        assertEquals(compb_scopeA.a, compb_scopeB.a)

        stopKoin()
    }
```

just pushed it for alpha-10

NkD:
@arnaudgiuliani ok, this seems good, but....
Using a linkTo has some limitations, I think.
In my opinion, it would be better to allow the creation of hierarchical Koin rather than allow link between scopes.

arnaudgiuliani:
what limitation do you see in Scope Links?

NkD:
@arnaudgiuliani can be scope inside scope? I think no. This first limitation. Small but there is. I can create more "child" scopes on same level, but it will be confused a little bit.
Complete Koin chierarchy (parent/child) is more well-arranged for me then scope hierarchy.
But I do not want to complain, because even scopes hierachy I can do what I need.

nikiJava:
> Hello @NkD
> 
> thanks for your submission. Having Koin modules spread around Gradle modules is already something you can do: https://doc.insert-koin.io/#/koin-core/modules?id=using-several-modules
> 
> Why do you need to have separated Koin instance for your modules then? 🤔

Excuse me, but what is the best practice when use koin in multi module project?
Make load/unload koin-modules in every gradle module or use Scope API and create scope for every gradle module?
I just wanna understand what is the reason to use scopes if it's just enough to use modules only?

arnaudgiuliani:
For multiple gradle module app, just use multiple Koin modules. You can even load/drop instances loaded in a module.

Scope deals with the problem of space where you want to resolve some temporary instances (else it's just singletons)

Using separated Koin instances is interesting if you want to isolate your components from any other Koin instance.

I still don't really why you would have a same app using its modules with seperated Koin instances, and make them communicate ... just use modules 🤔

arnaudgiuliani:
The execution will remain in a final same Koin instance in your story ...

PR #719, opened 2020-02-16
Workmanager support

fmatosqg:
See AppModule.kt on how to create a scoped module that is able to instantiate an object WorkerListenable (aka WorkModule framework), with extra constructor parameters provided by koin itself.

See also WorkManagerActivity  for a working activity sample.


fmatosqg:
@arnaudgiuliani do you see this PR as relevant?

arnaudgiuliani:
Hey @fmatosqg 

Interesting contribution, thanks.  Why do we need a `workerModule`? 🤔

If a WorkManager component needs a special factory, we can propose a new DSL keyword. The approach seems very similar to `ViewModel`?

fmatosqg:
I don't remember things in detail right now and can't check in detail, but if the approach for view model works, by all means.

I think there's a significant difference in how work manager factory works because some of the parameters are created just before your worker is instantiated. Maybe the scenario where work b uses the result of work a would illustrate this better. I think that's where I create and use a koin scope. There's a unit test, it may be more accurate than my words here, it's been a while.

arnaudgiuliani:
The main entry point is, is there any dedicated factory to use? Can you send back some documentation entry point to help identify how we can help with the manager globally?

For Android components such as ViewModel, the key is that we need to plug on to the factory API to help make the instance creation.

fmatosqg:
This is an official post from the google team with details about how to instantiate `ListenableWorker` with extra parameters:  https://medium.com/androiddevelopers/customizing-workmanager-fundamentals-fdaa17c46dd2

The general idea is that `ListenableWorker` is an abstract class that when extended runs the work that the end user wants to have performed. It needs 2 params, `Context` and `WorkerParameters`, plus extra constructor params that the end user wants. 

The difference here with `ViewModel` factories is that the only place where you can find an instance of `WorkerParameters` is inside a `WorkerFactory::createWorker(Context,String,WorkerParameters):ListenableWorker`.  And that's where the solution with a custom scope comes in: the scope captures `WorkerParameters` and allows us to make a "typical" koin factory and inject it with `get()`.

After the factory is written, the last part is how to make the app use it: call initialize from application and disable default initialization (because if called 2x it throws an exception).  https://developer.android.com/reference/kotlin/androidx/work/WorkManager#initialize(android.content.Context,%20androidx.work.Configuration)

arnaudgiuliani:
The key for `ViewModel` API is to wrap it as a Koin `factory` and give this instance to the right Android ViewModel factory when needed. The `by viewModel()` unlock the instance from the Koin factory, and pass it to the ViewModel factory.

Wouldn't it be similar as we would provide a `workManager` DSL key word to help declare a component, allowing its constructor injection.

From Activity/Fragment, having a `by workManager()` to create instance and pass it to the right Factory. We can also help compose/write factory on the fly.

Interesting, let me check/read your mentioned doc  :)

fmatosqg:
In that regard workers are different, since they're not instantiated or even accessible in typical user written classes like fragments/presenters. Being asynchronous by nature, such class would declare that it wants some work performed and mentions the worker through generics or `String`. Only when the adequate conditions are met (like phone is charging, or wifi is available, etc) is that the factory will kick in, instantiate the worker, and hand it over to the framework logic who will be in charge of running it and checking whether it finished successfully. 

For example, an activiy (koin/sample/androidx/workmanager/WorkManagerActivity.kt ) could write this on `onCreate`
```
OneTimeWorkRequestBuilder<Worker1>().build
.let{ 
  ....enqueueUniqueWork(....,it)
```

So unless the user writes the factory himself, I don't see a place for a dsl such as `by workManager()`. And in such case, if you write the factory then you can easily instantiate such object by using something like `return MyWorker(get<SomeDao>(), context, workerParameters)`. I'd be glad to be wrong, since this makes things more complex.

Here's where I stepped in and wrote in the same module a factory smart enough to resolve the name of the desired module and make an instance of it. For easier reference, a quick implementation would look someting like this
```
 override fun createWorker(
        appContext: Context,
        workerClassName: String,
        workerParameters: WorkerParameters
    ): ListenableWorker? {

   when(workerClassName){
     "Worker1" -> Worker1(get(),context,workerParameters)
     "Worker2" -> Worker2(get(),context,workerParameters)
     else -> null
  }
}
```

Here's where the most interesting part comes in. It replaces the `when` with a  `Map<String,lambda>`, or to be more specific, a `WorkerFactoryCreatorBlock`

```
typealias WorkerFactoryCreatorBlock = Function2<Context, WorkerParameters, ListenableWorker> ```

and to create such lambdas you simply have in a koin module something like this

```
 workerModule {
    worker { Worker1(get(), get(), get()) }
```

That's a quick rundown of what the code is supposed to do. I hope it makes sense, but certainly having a look at medium link in previous comment will be a better place to start.

**Comment by arnaudgiuliani** [comment]
Ok, thanks for your explanation. 

I didn't get why you need `workerModule`?

**Comment by fmatosqg** [comment]
Ideally we would simply use a factory / module
```
module {
  worker { DummyWorker(get(), get(), get< WorkerParameters>()) }
}
```

but a module wouldn't be able to resolve `get<WorkerParameters>()`. Maybe there's an alternative way to do that, I'm not proficient in the internals of the lib.
=====================
The next natural question would be why we need a `worker( lambda )` dsl. The reason is that this lambda is going to get stored in `AbstractWorkerFactory::addCreator` on  `mutableMapOf<String?, WorkerFactoryCreatorBlock>()`, and run on `AbstractWorkerFactory::createWorker` when the app needs to run such work. 

At "dependency tree building time" `KoinWorkerFactory::createWorker` creates the lambda stored on the map. Inside this lambda we (1) create a scope with  `getOrCreateScope` and give it an object of `params:WorkerParams` with `declare(params)`; then (2) invoke the lambda from `ScopeDSLExt.kt`, who will (3) invoke the lambda from `AppModule.kt`. 

(2) has access to the scope created in (1), so it can resolve basically any parameters needed, included but not limited to `WorkerParams`.

At the moment the worker is going to be run the framework makes a call to `AbstractWorkerFactory::createWorker`, who has a hold on a `params: WorkerParams` object. Then it invokes (1), (2) and (3).

Here's where there may be room for improvement since it was kinda tricky for me to get this right.

Going over the code to explain this I noticed that actually  `KoinWorkerFactory::createWorker` should be called `KoinWorkerFactory::addWorker` instead.

fmatosqg:
@arnaudgiuliani pls hold merging this PR for now. I found a bug, and changed a couple of things that made it much simpler, but I still need to polish my last commit to make sure of one particular case.

arnaudgiuliani:
@fmatosqg I finally got time to give your more details.

I believe the approach have to be more simplified and "in the way of Koin". 

One starting point to understand, is that we have the same problem of injecting state with `StateViewModelFactory` - https://github.com/InsertKoinIO/koin/pull/856/files#diff-551c6044a4051a4b761fac1895eab2b1
The `StateDefinitionParameter` allows us to validate an injected parameter, for Bundle state purpose here. https://github.com/InsertKoinIO/koin/pull/856/files#diff-7ae7925af0f5d0fed865c1e9b2bc397c
The `VireModel` DSL keyword allow us to redefine a Koin factory definition (Koin don't need to hold the instance) - 

What we need:
- a `KoinWorkerFactory`, that will use the root scope of Koin to resolve a Worker definition. Like the KoinFragmentFactory, this factory will be setup before (unlike ViewModel, where we provide one factory per instance on the fly). As you mentionned, we need to declare it in the manifest (if I follow)

Same as StateViewModel approach, we just need a DSL keyword to help us redefine a factory definition: this instance created by Koin will be forwarded to the given Factory. And we will handle the injected parameters, to inject the Worker paramerters. This is done via `ParametersDefinition` component.
 
Here I believe that your reinvent a factory over Koin, with a complex approach.

fmatosqg:
@arnaudgiuliani Reworked lots of things, it resembles a lot more the approach for fragment and view model, which is definitely easier than my initial approach.

One place where I got stuck was understanding how to use `ParametersDefinition`. For view model states, it seems to be necessary to use a syntax different than `get()`.

as in (a) `worker { (workParams, context) SomeWorker(workParams, context, get()) }`
as opposed to a more natural (b) `worker { SomeWorker(get(), get(), get()) }`.

If (a) is the way to go I'll have a second look, but I tried to make (b) work with the available tools.

And the other thing which it's only fair to say very controversial, is the way I handled it with scopes, and specially, copying the definition of SomeWorker from root scope into the factory scope. To avoid that I though about 1- linking scopes, but in this case I'd need to link Root -> workerScope, which is not allowed. I could 2- add `workParams` to root scope, but that would make collisions across different `worker { SomeWorker...` invocations. I also tried 3- doing some sort of ` scope<KoinWorkerFactory> {  factory { SomeWorker ` but couldn't find the way to accomplish what I was trying.

`.md` file is still in progress until we can work out the dsl.

arnaudgiuliani:
Ok, let me check again your code.

Just added the capacity to inject parameters into the graph: https://github.com/InsertKoinIO/koin/pull/889

arnaudgiuliani:
I can rework a branch from your work here, and integrate it in 2.2.x

fmatosqg:
Oh that would be amazing.

Btw only now I saw message about #889, will have a look at it later.

arnaudgiuliani:
I'm pushing your PR to the work_manager branch. I'll push things on it. I keep you i ntouch

PR #722, opened 2020-02-19
use actual filename in androidFileProperties

elexx:
instead of the hardcoded default filename, errormessages use the actual
filename that was not found or could not be loaded


arnaudgiuliani:
thanks

PR #735, opened 2020-02-27
Updated dependency quote marks in setup docs

aruke:
The dependency declaration in the docs uses the `koin_version` variable and is supposed to be replaced by a predefined version of Koin.
In some places, single-quote `'` is used with interpolation, which Groovy does not support.
For example:
```
'org.koin:koin-android:$koin_version'
```
Direct copy-paste dependencies from docs will cause sync errors in Gradle files.

Groovy supports string interpolation in double quotes `"`, so I replaced the single quotes with double.

[Groovy Syntax Reference](https://groovy-lang.org/syntax.html) `4.4.1. String interpolation`

### DIFF
--- a/koin-projects/docs/start/quickstart/android-java.md
+++ b/koin-projects/docs/start/quickstart/android-java.md
@@ -19,7 +19,7 @@ repositories {
 }
 dependencies {
     // Koin for Android
-    compile 'org.koin:koin-android:$koin_version'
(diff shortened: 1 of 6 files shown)


aruke:
@1jGabriel Thanks for the approval. Whom should I connect to merge this PR?

arnaudgiuliani:
@aruke thanks for your contribution

PR #742, opened 2020-03-05
Add fallback for instantiate function in FragmentFactory Fixes gh-738

### LINKED ISSUE #738: Instantiate fallback in KoinFragmentFactory
I'm using `FragmentFactory` in my application, and i found that some fragments are instantiated by framework or libraries themselves. The only way to find out about it is to get crash at some point. Two examples i can give is `NavHostFragment` from Jetpack Navigation or `SupportRequestManagerFragment` from `Glide` library (this was especially tricky. I found out about this fragment from crashlytics raport).

Maybe if we can't find proper instance we should use fallback instantiate from super class? `super.instantiate(classLoader, className)`. This way - if your fragment have some parameters app will fail fast. But fragments with empty constructors will get created "the old way". What do you think?

**Target Koin project**
androix.FragmentFactory

Nekromancer:
With this fix fragments with empty constructors will get instantiated "the old way" when not registered in koin container. This is helpful for fragments that does not require any parameters (in my case - all of external libraries fragments).

If user will define fragment with some parameters - instantiate will still fail, so there is no danger for missuse here

Resolves #738


Nanamare:
nice :)

PR #773, opened 2020-04-10
support declare a component definition with primary type

xudshen:
# Problem

When we want to declare an instance on the fly, the current logic uses the instance's class type as primary type for the bean definition.
But there are sometimes that we might expect using instance's supertype as primary type.

```kotlin
open class Lifecycle
class UserLifecycle1 : Lifecycle()
class UserLifecycle2 : Lifecycle()

val koin = koinApplication {
    printLogger()
    modules(module {
        scope(named("UserSession1")) {
            scoped(named("lifecycle")) { UserLifecycle1() as Lifecycle }
        }
        scope(named("UserSession2")) {
        }
    })
}.koin

val session1 = koin.createScope("session1", named("UserSession1"))
val session2 = koin.createScope("session2", named("UserSession2"))
session2.declare(UserLifecycle2() as Lifecycle, qualifier = named("lifecycle"))

session1.get<Lifecycle>(named("lifecycle")) // return UserLifecycle1 instance
session2.get<UserLifecycle2>(named("lifecycle")) // return UserLifecycle2 instance
session2.get<Lifecycle>(named("lifecycle")) // throw definition not found
```

`scoped(named("lifecycle")) { UserLifecycle1() as Lifecycle }` and `session2.declare(UserLifecycle2() as Lifecycle, qualifier = named("lifecycle"))` should achieve the same effect intuitively.

# Solution

I've changed the `Scope#declare` function and added a `clazz` parameter to specify the primary type for current definition. Inside `ScopeDefinition#saveNewDefinition`, it will check if `clazz` is same as or supertype of provided `instance`.

```kotlin
// session2.declare(UserLifecycle2() as Lifecycle, qualifier = named("lifecycle"))
session2.declare(UserLifecycle2(), Lifecycle::class, qualifier = named("lifecycle"))

session2.get<Lifecycle>(named("lifecycle")) // return UserLifecycle2 instance as expected
```

### DIFF
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
@@ -139,13 +139,12 @@ class Koin {
      */
     inline fun <reified T : Any> declare(
         instance: T,
-        clazz: KClass<*> = instance::class,
         qualifier: Qualifier? = null,
         secondaryTypes: List<KClass<*>> = emptyList(),
         override: Boolean = false
     ) {
         val firstType = listOf(T::class)
-        _scopeRegistry.rootScope.declare(instance, clazz, qualifier, firstType + secondaryTypes, override)
+        _scopeRegistry.rootScope.declare(instance, qualifier, firstType + secondaryTypes, override)
     }
 
     /**
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
@@ -139,12 +139,13 @@ class Koin {
      */
     inline fun <reified T : Any> declare(
         instance: T,
+        clazz: KClass<*> = instance::class,
         qualifier: Qualifier? = null,
         secondaryTypes: List<KClass<*>> = emptyList(),
         override: Boolean = false
     ) {
         val firstType = listOf(T::class)
-        _scopeRegistry.rootScope.declare(instance, qualifier, firstType + secondaryTypes, override)
+        _scopeRegistry.rootScope.declare(instance, clazz, qualifier, firstType + secondaryTypes, override)
     }
 
     /**
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/scope/Scope.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/scope/Scope.kt
@@ -260,19 +260,17 @@ data class Scope(
      * (single definition of th current scope is root)
      *
      * @param instance The instance you're declaring.
-     * @param clazz The primary type for this declaration
      * @param qualifier Qualifier for this declaration
      * @param secondaryTypes List of secondary bound types
      * @param override Allows to override a previous declaration of the same type (default to false).
      */
-    fun <T : Any> declare(
+    inline fun <reified T : Any> declare(
         instance: T,
-        clazz: KClass<*> = instance::class,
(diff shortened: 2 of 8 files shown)


koin-projects/koin-core/src/main/kotlin/org/koin/core/scope/Scope.kt
```diff
@@ -260,17 +260,19 @@ data class Scope(
      * (single definition of th current scope is root)
      *
      * @param instance The instance you're declaring.
+     * @param clazz The primary type for this declaration
      * @param qualifier Qualifier for this declaration
      * @param secondaryTypes List of secondary bound types
      * @param override Allows to override a previous declaration of the same type (default to false).
      */
     fun <T : Any> declare(
```
arnaudgiuliani, line 268: Instead of having to provide a new class, we can directly rely on reified type: `inline fun <reified T : Any>`

koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
```diff
@@ -139,12 +139,13 @@ class Koin {
      */
     inline fun <reified T : Any> declare(
         instance: T,
+        clazz: KClass<*> = instance::class,
```
arnaudgiuliani, line 142: not sure to add it directly here, I would recommend a secondary function

koin-projects/koin-core/src/main/kotlin/org/koin/core/scope/ScopeDefinition.kt
```diff
@@ -36,11 +37,14 @@ class ScopeDefinition(val qualifier: Qualifier, val isRoot: Boolean = false, pri
...
             instance: T,
+            clazz: KClass<*> = instance::class,
             qualifier: Qualifier? = null,
             secondaryTypes: List<KClass<*>>? = null,
             override: Boolean = false
     ): BeanDefinition<out Any?> {
-        val clazz = instance::class
+        if (!clazz.isInstance(instance)) {
+            throw BadDefinitionException("Primary type '$clazz' should be or supertype of '${instance::class}'")
+        }
```
arnaudgiuliani, line 47: with reified type we could get rid of that, and avoid to pass a class

koin-projects/koin-test/src/main/kotlin/org/koin/test/mock/Declare.kt
```diff
@@ -22,6 +22,6 @@ inline fun <reified T : Any> KoinTest.declare(
         noinline instance: () -> T
 ): T {
     val koin = KoinContextHandler.get()
-    koin.declare(instance(), qualifier, override = true)
+    koin.declare(instance(), qualifier = qualifier, override = true)
```
arnaudgiuliani, line 25: necessary?

koin-projects/koin-test/src/main/kotlin/org/koin/test/mock/DeclareMock.kt
```diff
@@ -58,7 +58,7 @@ inline fun <reified T : Any> Scope.declareMock(
         stubbing: StubFunction<T> = {}
 ): T {
     val mock = MockProvider.makeMock<T>()
-    declare(mock, qualifier, secondaryTypes + T::class, true)
+    declare(mock, qualifier = qualifier, secondaryTypes = secondaryTypes + T::class, override = true)
```
arnaudgiuliani, line 61: necessary?

xudshen:
hello, I've changed the approach with reified type parameters. The only downside is making the `Definitions#createSingle` public.

PR #781, opened 2020-04-17
Disable property type cast to fix issue gh-291

### LINKED ISSUE #291: Number-like strings in koin.properties
(This is somewhere between a feature request and a bug report...)

There doesn't seem to be a way to get a string value from `koin.properties` when it looks like an `Int` or `Float`. For example, if `koin.properties` has:
```
OAUTH_APP_ID=1234567
```
then using
```
val appId: String by property("OAUTH_APP_ID")
```
throws an exception, because `PropertyRegistry.import` has added it as the detected `Int` and the cast in `getValue<T>` blows up.

I'd like to have some mechanism to get a number-like string value from `koin.properties` (or, for that matter, system environment).

Surrounding the value with quotes, eg. `OAUTH_APP_ID="1234567"` keeps the quotes as part of the parsed value. However, I'm not sure that would resolve the issue when pulling values from the system environment.

Since any value should be able to be represented by a string, perhaps `PropertyRegistry.getValue<T>` could be altered to return the type guessed during import OR a String when that's what is requested?

**Target Koin project**
`koin-core`

**Edit**
This cannot be worked around when the value being parsed overflows. `OAUTH_APP_ID=386226214594673` will end up giving the value `386226209161216`, and no amount of casting will fix that.

We also see this limited to `Int` and `Float` but not for other basic types like `Boolean`. Perhaps all values should be imported as strings and the casts are only performed with plucking the values out?

Swordsman-Inaction:
Fixes #291 

Two changes(Breaking changes):

- Disable property type cast, now property value can only be `String`
- Add extension functions for `int` and `float` properties.

Before:
- `getKoin().getProperty<Int>("int.value")`
- `getKoin().getProperty<Float>("float.value")`

Now:
- `getKoin().getIntProperty("int.value")`
- `getKoin().getFloatProperty("float.value")`


koin-projects/koin-core/src/main/kotlin/org/koin/ext/KoinPropertyExt.kt
```diff
@@ -0,0 +1,63 @@
+package org.koin.ext
```
arnaudgiuliani, line 1: Just renamed it as `KoinPropertyExt.kt`

arnaudgiuliani:
Please also check conflicts

koin-projects/koin-core/src/main/kotlin/org/koin/ext/KoinPropertyExt.kt
Swordsman-Inaction: Sure

Swordsman-Inaction:
@arnaudgiuliani  Updated and tested.
But the `MVVMActivity` in `androidx-samples` project is crashing because of this change: https://github.com/InsertKoinIO/koin/commit/53c07fbcf0a01713e6ed14c059c35e8b695aed2a#diff-2263db8ed8a57c624f749b4114746255R45

jeziellago:
> @arnaudgiuliani Updated and tested.
> But the `MVVMActivity` in `androidx-samples` project is crashing because of this change: [53c07fb#diff-2263db8ed8a57c624f749b4114746255R45](https://github.com/InsertKoinIO/koin/commit/53c07fbcf0a01713e6ed14c059c35e8b695aed2a#diff-2263db8ed8a57c624f749b4114746255R45)

#795 should fix `androidx-samples`. 
https://github.com/InsertKoinIO/koin/pull/795/files#diff-2263db8ed8a57c624f749b4114746255R53

arnaudgiuliani:
Ok, will check that asap 👍

arnaudgiuliani:
just merged the PR

arnaudgiuliani:
can you double check @Swordsman-Inaction ? 🙏

Swordsman-Inaction:
@arnaudgiuliani Just merged `2.1.x`, all good now.

chrisjenx:
Don't wanna be that guy - but yo - this really does not go into a _bugfix_ release! this is a Major a release at best, Minor release at worst. If you are going to use semver at least respect it!

Satook:
I agree with @chrisjenx. This is a backward incompatible change. I'm stuck on 2.1.5 until I change how I pass config data around my app.

FYI: I'm using Ktor which has a richer ApplicationConfig (it supports nesting and lists of values). Means I'll have to stop using properties and move it over to an injectable parameter.

Would a patch that keeps the same external interface (getProperty, getIntProperty, getFloatProperty) be accepted? I'd add getAnyProperty or getTypedProperty<T> and swap the map back to <string, any>. This way we can keep the string oriented interfaces for people who just want a flat PropertyList style but give others in my situation a good migration path.

koin-projects/koin-core/src/main/kotlin/org/koin/ext/KoinPropertyExt.kt
antohaby, line 1: This file isn't complete enough.
On the previous version `getProperty` was accessible from KoinScope as well.

For example: 

```
single {
    ConnectionPool (
        maxConnections = getIntProperty("CONNECTION_POOL__MAX_CONNECTIONS")
    )
}
```

is not going to work, but should IMHO.

Issue #763, opened 2020-03-28
Koin-test support for Junit 5


GregHib:
**Is your feature request related to a problem? Please describe.**
The exclusive use of Rules in `koin-test` forces users to use JUnit 4 with no apparent alternative for JUnit 5 users.

**Describe the solution you'd like**
Koin-test should provide JUnit 5 extensions. I presume there are other JUnit 4 usages which might also need migrating (e.g `AutoCloseKoinTest.kt`).

## Code
`KoinTestRule.kt` - `KoinTestExtension.kt`
```kotlin
/**
 * [Extension] which will automatically start and stop Koin.
 * @author Nick Cipollo
 * @author Greg Hibberd
 */
class KoinTestExtension private constructor(private val appDeclaration: KoinAppDeclaration) : BeforeEachCallback, AfterEachCallback {

    var _koin: Koin? = null
    val koin: Koin
        get() = _koin ?: error("No Koin application found")

    override fun beforeEach(context: ExtensionContext) {
        _koin = startKoin(appDeclaration = appDeclaration).koin
    }

    override fun afterEach(context: ExtensionContext?) {
        stopKoin()
        _koin = null
    }

    companion object {
        fun create(appDeclaration: KoinAppDeclaration): KoinTestExtension {
            return KoinTestExtension(appDeclaration)
        }
    }
}
```

`MockProviderRule.kt` - `MockProviderExtension.kt`
```kotlin
/**
 * @author Greg Hibberd
 * @since March 28, 2020
 */
class MockProviderExtension private constructor(private val mockProvider: Provider<*>) : BeforeEachCallback {

    override fun beforeEach(context: ExtensionContext) {
        MockProvider.register(mockProvider)
    }

    companion object {
        fun create(mockProvider: Provider<*>): MockProviderExtension {
            return MockProviderExtension(mockProvider)
        }
    }
}
```

## Usage
```kotlin
@JvmField
@RegisterExtension
val koinTestExtension = KoinTestExtension.create {
    TODO("Implement")
}

@JvmField
@RegisterExtension
val mockProvider = MockProviderExtension.create { clazz ->
    TODO("Implement")
}
```

gabin8:
junit 5 support would be very helpful

niom:
As far as I have understood the whole junit5 Extensions. They should not have variables that are mutated from the callbacks but instead of mutating they should put the state inside the ExtensionContext store:

```
//store
[...]

PR #784, opened 2020-04-19
Feature/junit5 support

niom:
This PR will add support for junit5: #763 

Support junit5 with extension and also AutoCloseKoinTest. Both can be found from package path: org.koin.test.junit5.

old AutoCloseKoinTest can be found from module koin-test-junit4 and is still at the same package path as it was before. Only the module isa moved

This modification contains breaking changes considering module dependencies since junit4 and junit5 test code needs to be in separate modules. There for using koin-test dependency does not include the Rule or the AutoCloseKoinTest functionality. 

Now users need to add either koin-test-junit4 or koin-test-junit5 dependency in to their project to get the correct AutoClosing functionality or Rule/Extension functionality. Backwards compatibility is still ensured when users will add koin-test-junit4 dependency into their dependency tree. 

Testing is done mainly using the same test functionality at the junit4 side usign copy-paste and modifying the tests to use the new junit5 functionality. Also wrote tests that verify that KoinTestExtension and KoinTestRule works as expected.

Do you require a README.md or similar for how to use this functionality so users can easily check how junit5 testing is done with koin?


arnaudgiuliani:
Thanks for you contribution. Interesting work.

Don't know if we have to maintain JUnit 4 & 5 ... does JUnit 4 is deprecating somewhere?

GregHib:
> 
> 
> Thanks for you contribution. Interesting work.
> 
> Don't know if we have to maintain JUnit 4 & 5 ... does JUnit 4 is deprecating somewhere?

I can't find any reference to JUnit 4 being deprecated however looking at the code frequency of the projects it hasn't been in active development since JUnit5 started in 2015

niom:
We'll at least have to have the different rule / extension for them. So maybe in this case the separation is unnecessary. I'll check if we can live with rules using the vintage junit and then the extensions using the junit5 directly

arnaudgiuliani:
Yes, good catch 👍

igorwojda:
AFFAIK there are no plans for depreciating JUnit4 for Android, however frameworks like [Spek](https://www.spekframework.org/) are already build on top of JUnit5.

I think it is worth supporting JUni5 since it looks like these versions does not conflict with each other. Actually JUnit4 `Runner`, `TestRule` / `MethodRule` evolved in JUnit5 into single concept called `Extension`

arnaudgiuliani:
Ok, interesting to have it in the tools then. Don't if it's 2.1.6 or 2.2.0 target. But definitively ok to adopt it 👍

arnaudgiuliani:
Hi,

can you rebase your contribution on 2.2.0? There is conflict

Thanks

arnaudgiuliani:
Last comment on your proposal, I prefer that we keep `koin-test` as the default module with JUnit4 for now. Your contribution can bring the `koin-test-junit5` one.

niom:
conflict resolved. 

I needed to do some module renaming based on the requirement to keep the junit4 as the default testing library.
There's still going to be three separate libraries for the koin test functionality 
koin-test-core for shared functionality
koin-test that inherits the koin-test-core as a dependency and provides the junit4 testing functionality
koin-test-junit5 that inherits the koin-test-core as a dependency and provides the junit5 testing functionality.

With all the time that this PR has been pending on my side I gave it a lot of thought about using the vintage junit5 driver for junit4. I also tried it briefly months ago and It would have been possible to replace the junit4 with the vintage driver but would it been the the correct way to force projects to have the vintage dependency. It felt kind of weird. So my conclusion is the same as with you that to stick with the junit4 since it's not deprecated.

arnaudgiuliani:
> conflict resolved.
> 
> I needed to do some module renaming based on the requirement to keep the junit4 as the default testing library.
> There's still going to be three separate libraries for the koin test functionality
> koin-test-core for shared functionality
> koin-test that inherits the koin-test-core as a dependency and provides the junit4 testing functionality
> koin-test-junit5 that inherits the koin-test-core as a dependency and provides the junit5 testing functionality.
> 
> With all the time that this PR has been pending on my side I gave it a lot of thought about using the vintage junit5 driver for junit4. I also tried it briefly months ago and It would have been possible to replace the junit4 with the vintage driver but would it been the the correct way to force projects to have the vintage dependency. It felt kind of weird. So my conclusion is the same as with you that to stick with the junit4 since it's not deprecated.

great 👍 

lastly, we need just a bit of doc if you can add some explanation about what you introduced. Add just a new section about JUnit5 tools in `./docs/reference/koin-test/testing.md`

Its just markdown.

Add something like 

```
## Testing with JUnit5
...
```

I will help you with that 👍 , thanks

niom:
I added documentation how to use the JUnit5 extension. 

One thing that I was thinking but did not add was how to add the JUnit5 dependency. There's a header for that but I did not include any maven or gradle configuration instructions. Should I add that Or should it be added to an other md file?

arnaudgiuliani:
Let's go 👍 

Thanks for your work

navi89mai:
@arnaudgiuliani any plan to release new version?

I see two reasons to release: junit5 support, kotlin 1.4

PR #790, opened 2020-04-22
share common scope

softartdev:
The case of MVP architecture, when a common interactor is shared between presenters.
The purpose is to reuse an already created interactor if another screen with the same functions has open.

### DIFF
--- a/koin-projects/examples/coffee-maker/src/test/kotlin/org/koin/example/SharedScopeTest.kt
+++ b/koin-projects/examples/coffee-maker/src/test/kotlin/org/koin/example/SharedScopeTest.kt
@@ -1,0 +1,92 @@
+package org.koin.example
+
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotEquals
+import org.junit.Rule
+import org.junit.Test
+import org.koin.core.logger.Level
+import org.koin.core.qualifier.named
+import org.koin.core.scope.Scope
+import org.koin.dsl.module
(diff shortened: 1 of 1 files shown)


arnaudgiuliani:
Why such proposal on coffee-maker example?

softartdev:
> Why such proposal on coffee-maker example?

Such an example used by sample project, so it doesn’t matter, the main point of this pull request is different.
I need something like [reusable scope annotation from Dagger](https://dagger.dev/api/2.21/dagger/Reusable.html). 
So far I have not figured out: does the current version of Koin have such an opportunity or should it be expected in the future? 
And if it is already possible, how to do it right? Because the way it is written in my example does not seem right to me. If DI-things (in this case, scope) mentioned in presenters, then Single Responsibility Principle is broken. I tried to write a call to [scope.close()](https://github.com/InsertKoinIO/koin/blob/c86475bf5517920f38d3bfb98f2f69d9a2303453/koin-projects/examples/coffee-maker/src/test/kotlin/org/koin/example/SharedScopeTest.kt#L33) inside the [module](https://github.com/InsertKoinIO/koin/blob/c86475bf5517920f38d3bfb98f2f69d9a2303453/koin-projects/examples/coffee-maker/src/test/kotlin/org/koin/example/SharedScopeTest.kt#L58), but failed.

arnaudgiuliani:
Closing master branch closed all PRs :/ really sorry. Can you update your PR against main branch

PR #795, opened 2020-05-02
Fixing issue #787

### LINKED ISSUE #787: stateViewModel(bundle = intent.extras) can't be used at all?
**To Reproduce**
Koin version: `2.1.5`

When using Koin to inject ViewModel with savedStateHandle as below, all works well.
```
private val viewModel : MyViewModel by stateViewModel()
```
However, if I wanted to pass in the intent.extra, it crashes complaining the intent is null.
```
private val viewModel : MyViewModel by stateViewModel(bundle = intent.extras)
```

This is not ideal, meaning whenever we have `bundle` to pass in, we can't use `stateViewModel`.

We have to workaround using
```
private val viewModel by lazy {
    getStateViewModel<MyViewModel> (bundle = intent.extras)
}
```

Refer to below for more info
https://stackoverflow.com/questions/61224797/stateviewmodel-injection-on-koin-crash-when-sending-intent-extra-parameter

jeziellago:
This PR fix #787. 

Instead of (crashing):
```kotlin
private val viewModel : MyViewModel by stateViewModel(bundle = intent.extras)
```

Replace bundle to:
```kotlin
private val viewModel : MyViewModel by stateViewModel(bundle = { intent.extras })
```

### DIFF
--- a/koin-projects/koin-androidx-viewmodel/src/main/java/org/koin/androidx/viewmodel/ViewModelParameter.kt
+++ b/koin-projects/koin-androidx-viewmodel/src/main/java/org/koin/androidx/viewmodel/ViewModelParameter.kt
@@ -15,3 +15,5 @@ class ViewModelParameter<T : Any>(
         val viewModelStore: ViewModelStore,
         val stateRegistryOwner: SavedStateRegistryOwner? = null
 )
+
+typealias StateBundleParameter = () -> Bundle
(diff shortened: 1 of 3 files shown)


jeziellago:
> Thanks for your contribution,
> 
> can you update existing androidx-example project to update tests?

Done! 😁

PR #822, opened 2020-06-05
keep createSingle internal

arttttt:


arnaudgiuliani:
Hi @arttttt 

thanks for the proposal. I would better consider a whole pass on cleaning API visibility

PR #843, opened 2020-07-01
Better checkModules API with default values & Mocked dependencies

arnaudgiuliani:
Hi all,

here is a PR to fix `checkModules` API with the following:

- bring default values and Mock to help verify injected parameters definition without having to define a `create` for each definition
- allow to inject default Scope instance origin value

This should clearly help `checkModules` experience, as you can mainly check your modules without much extra setup.

### DIFF
--- a/koin-projects/examples/androidx-samples/src/main/java/org/koin/sample/androidx/components/scope/Session.kt
+++ b/koin-projects/examples/androidx-samples/src/main/java/org/koin/sample/androidx/components/scope/Session.kt
@@ -1,7 +1,7 @@
 package org.koin.sample.androidx.components.scope
 
 import org.koin.sample.androidx.scope.ScopedActivityA
-import java.util.*
+import java.util.UUID
 
 data class Session(var id: String = UUID.randomUUID().toString())
 
--- a/koin-projects/examples/androidx-samples/src/main/java/org/koin/sample/androidx/components/scope/Session.kt
+++ b/koin-projects/examples/androidx-samples/src/main/java/org/koin/sample/androidx/components/scope/Session.kt
@@ -1,8 +1,8 @@
 package org.koin.sample.androidx.components.scope
 
-import android.app.Activity
-import java.util.UUID
+import org.koin.sample.androidx.scope.ScopedActivityA
+import java.util.*
 
 data class Session(var id: String = UUID.randomUUID().toString())
 
-data class SessionActivity(val activity: Activity, val id: String = UUID.randomUUID().toString())
+data class SessionActivity(val activity: ScopedActivityA, val id: String = UUID.randomUUID().toString())
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
@@ -122,9 +122,9 @@ class Koin {
      * @return instance of type T or null
      */
     fun <T> getOrNull(
-            clazz: KClass<*>,
-            qualifier: Qualifier? = null,
-            parameters: ParametersDefinition? = null
+        clazz: KClass<*>,
+        qualifier: Qualifier? = null,
+        parameters: ParametersDefinition? = null
     ): T? = _scopeRegistry.rootScope.getOrNull(clazz, qualifier, parameters)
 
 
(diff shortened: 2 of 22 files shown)


koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
```diff
@@ -228,9 +228,10 @@ class Koin {
      * Get or Create a Scope instance
      * @param scopeId
      * @param qualifier
+     * @param source
      */
-    fun getOrCreateScope(scopeId: ScopeID, qualifier: Qualifier): Scope {
-        return _scopeRegistry.getScopeOrNull(scopeId) ?: createScope(scopeId, qualifier)
+    fun getOrCreateScope(scopeId: ScopeID, qualifier: Qualifier, source: Any? = null): Scope {
```
kirich1409, line 233: For what the source param is responsible for?

koin-projects/examples/androidx-samples/src/main/java/org/koin/sample/androidx/components/scope/Session.kt
```diff
@@ -1,8 +1,8 @@
 package org.koin.sample.androidx.components.scope
 
-import android.app.Activity
-import java.util.UUID
+import org.koin.sample.androidx.scope.ScopedActivityA
+import java.util.*
```
ppjeziellago, line 4:
```suggestion
import java.util.UUID
```

koin-projects/koin-test/src/main/kotlin/org/koin/test/check/CheckModules.kt
```diff
@@ -46,42 +49,53 @@ fun checkModules(level: Level = Level.INFO, parameters: CheckParameters? = null,
...
-        bindings.parametersDefinition()
-    }
-    val allParameters = bindings.creators
-    return allParameters
+private fun Koin.declareParameterCreators(
+    parametersDefinition: CheckParameters?
+): ParametersBinding {
+    val bindings = ParametersBinding(this)
+    parametersDefinition?.let { bindings.parametersDefinition() }
+    return bindings
```
jeziellago, line 64:
```suggestion
) = ParametersBinding(this).also { binding -> 
    parametersDefinition?.invoke(binding) 
}
```

koin-projects/koin-test/src/main/kotlin/org/koin/test/parameter/MockParameter.kt
```diff
@@ -0,0 +1,28 @@
...
+            Int::class.java.simpleName -> 0 as T
+            Double::class.java.simpleName -> 0.0 as T
+            Float::class.java.simpleName -> 0.0f as T
+            else -> {
+                val found = try {
+                    scope.getOrNull<T?>(clazz = clazz)
+                } catch (e: Exception) {
+                    // not found
+                }
+                (found ?: Mockito.mock(clazz.java)) as T
```
jeziellago, line 24: `Scope.getOrNull` already has `try/catch` block and returns `null` if catches some exception.
You could simplify as:
```suggestion
                (scope.getOrNull<T?>(clazz = clazz) ?: Mockito.mock(clazz.java)) as T
```

koin-projects/koin-core/src/main/kotlin/org/koin/core/Koin.kt
arnaudgiuliani: source is the origin instance that can hosted by the scope. Like `scope<MyActivity>` will have an `MyActivity` instance.

koin-projects/koin-test/build.gradle
```diff
@@ -7,7 +7,7 @@ dependencies {
     // Koin
     compile project(":koin-core")
     compile "junit:junit:$junit_version"
-    testImplementation "org.mockito:mockito-inline:$mockito_version"
+    compile "org.mockito:mockito-inline:$mockito_version"
```
erickok, line 10: Forcing mockito is a too strong requirement IMHO. There are project that use different mocking frameworks - which might conflict - or do not want any mocking framework at all - for example due to build speed considerations.

arnaudgiuliani: Yes, I clearly need a way to mock needed injected params, as we don't need to write a parameter creator for each of them. It will help people scale their module with that.

Another thigns can be to backport just the mock part of a lib

arnaudgiuliani: I finally rework to use the `MockProvider` rule, that allows to provide mock with the desired framework

erickok: Much better. Still not the greatest fan of mocking by default (via `?: MockParameter()`) but at least you have to opt-in or you would get an error.

PR #857, opened 2020-07-30
Replace fragment extension

GabrielBrasileiro:
A simple way to recover and instantiate fragment. :)

I create this extension to make a sugar declaration when we need create the fragment. With this form we can set the tag only with the alias in method.

Before:
```kotlin
supportFragmentManager.beginTransaction()
    .replace(R.id.mvvm_frame, MVVMFragment::class.java, null, null)
    .commit()
```

After:
```kotlin
supportFragmentManager.beginTransaction()
      .replace<MVVMFragment>(R.id.mvvm_frame)
      .commit()
```

### DIFF
--- a/koin-projects/docs/reference/koin-android/fragment-factory.md
+++ b/koin-projects/docs/reference/koin-android/fragment-factory.md
@@ -65,10 +65,20 @@ And retrieve your `Fragment` with your `supportFragmentManager`:
 
 ```kotlin
 supportFragmentManager.beginTransaction()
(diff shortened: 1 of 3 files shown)


arnaudgiuliani:
Hi @GabrielBrasileiro 

interesting extension. Can you complete doc `docs/reference/koin-android/fragment-factory`?

Thanks :)

GabrielBrasileiro:
Of course yes!

PR #863, opened 2020-08-06
Add new create method using KClass on check-dsl for test

jeziellago:
## Description
This PR add a new create method with KClass argument instead of reified T type. 
In some cases, using KClass instead of T type directly can enable us to omit implementations classes and expose only its KClass for configuration of parameters in tests.

### Current
```kotlin
		...
		.checkModules {
			// Depends directly of MyViewModel and `user` object
            create<MyViewModel> { parametersOf(user) }
        }
```

### Expected (using KClass)
Could be used "stubs" providing only KClass and DefinitionParameters
```kotlin
		...
		.checkModules {
            create(TestStubs.myViewModelKClass())) { TestStubs.myViewModelParameters() }
        }
```


arnaudgiuliani:
intersting addition

Issue #871, opened 2020-08-17
After updating Koin 2.1.6 version I Got crash like.   java.lang.NoSuchMethodError: No virtual method elapsedNow()D in class Lkotlin/time/TimeMark; or its super classes (declaration of 'kotlin.time.TimeMark' appears ....


mcnarek:
java.lang.NoSuchMethodError: No virtual method elapsedNow()D in class Lkotlin/time/TimeMark; or its super classes (declaration of 'kotlin.time.TimeMark' appears in /data/app/my.app.id-IIFeKPZKJaPEihLziq4cTg==/base.apk!classes3.dex)
        at org.koin.core.time.MeasureKt.measureDuration(Measure.kt:36)
        at org.koin.core.KoinApplication.modules(KoinApplication.kt:60)
        at my.app.id.App$onCreate$1.invoke(App.kt:51)
        at my.app.id.App$onCreate$1.invoke(App.kt:23)
        at org.koin.core.context.ContextFunctionsKt.startKoin(ContextFunctions.kt:39)
        at org.koin.core.context.ContextFunctionsKt.startKoin$default(ContextFunctions.kt:35)
        at my.app.id.App.onCreate(App.kt:40)
        at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1182)
        at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6460)
        at android.app.ActivityThread.access$1300(ActivityThread.java:219)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1859)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)`

in App.kt (Application) class in onCreate method

```
`startKoin {
            // use AndroidLogger as Koin Logger - default Level.INFO
            androidLogger()

            // use the Android context given there
            androidContext(this@App)

            // load properties from assets/koin.properties file
            androidFileProperties()

            // module list
            modules(listOf(module1, module2, module3))
        }`
```

mcnarek:
Sorry this was after 1.4 update of Kotlin standard lib

[...]

Issue #847, opened 2020-07-09
[Crash] Koin Logger Measure and Kotlin EAP 1.4-M3


ParaskP7:
**Describe the bug**
This is just a heads up on `Koin` and `Kotlin EAP 1.4-M3`.

After upgrading to `Kotlin EAP 1.4-M3` (from previous `Kotlin EAP 1.4-M2`) when launching the app a `NoSuchMethodException` exception is thrown due to the `elapseNow()` experimental method.

Full exception below:
```
    java.lang.NoSuchMethodError: No virtual method elapsedNow()D in class Lkotlin/time/TimeMark; or its super classes (declaration of 'kotlin.time.TimeMark' appears in /data/app/io.petros.movies-u-dzv9wkMPKfx19uLVQWyg==/base.apk)
        at org.koin.core.time.MeasureKt.measureDuration(Measure.kt:36)
        at org.koin.core.KoinApplication.modules(KoinApplication.kt:60)
        at io.petros.movies.app.App$initKoin$1.invoke(App.kt:45)
        at io.petros.movies.app.App$initKoin$1.invoke(App.kt:23)
        at org.koin.core.context.ContextFunctionsKt.startKoin(ContextFunctions.kt:39)
        at org.koin.core.context.ContextFunctionsKt.startKoin$default(ContextFunctions.kt:35)
        at io.petros.movies.app.App.initKoin(App.kt:42)
        at io.petros.movies.app.App.onCreate(App.kt:34)
        at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1189)
        at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6460)
        at android.app.ActivityThread.access$1300(ActivityThread.java:219)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1859)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
```

[...]

PR #873, opened 2020-08-18
Fix Kotlin 1.4 runtime error on measure time

jeziellago:
Fixing runtime errors with kotlin 1.4  (#847 #871)

### DIFF
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/time/Measure.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/time/Measure.kt
@@ -15,9 +15,6 @@
  */
 package org.koin.core.time
 
-import kotlin.time.ExperimentalTime
-import kotlin.time.TimeSource
-
 
 /**
  * Measure functions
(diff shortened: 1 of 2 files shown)


arnaudgiuliani:
Before using this over deprecated API, I was just timing by myself. Do you see any good way without using an unstable API?

jeziellago:
> Before using this over deprecated API, I was just timing by myself. Do you see any good way without using an unstable API?

@arnaudgiuliani I rewrote without `kotlin.time`. Could you check it?

koin-projects/koin-core/src/main/kotlin/org/koin/core/time/Measure.kt
```diff
@@ -28,15 +25,12 @@ import kotlin.time.TimeSource
...
  * Measure code execution
  */
-@OptIn(ExperimentalTime::class)
 fun measureDuration(code: () -> Unit): Double {
-    val clock = TimeSource.Monotonic
-    val mark = clock.markNow()
+    val initialTime = System.nanoTime().toDouble()
     code()
-    return mark.elapsedNow().inMilliseconds
+    return (System.nanoTime() - initialTime).inMilliseconds()
```
arnaudgiuliani, line 31: just simple extract as `startTime` & `endTime` those variables 👍

koin-projects/koin-core/src/main/kotlin/org/koin/core/time/Measure.kt
```diff
@@ -45,17 +39,17 @@ fun measureDuration(message: String, code: () -> Unit) {
...
  * Measure code execution and get result
  */
-@OptIn(ExperimentalTime::class)
 fun <T> measureDurationForResult(code: () -> T): Pair<T, Double> {
-    val clock = TimeSource.Monotonic
-    val mark = clock.markNow()
+    val initialTime = System.nanoTime().toDouble()
     val result = code()
-    return Pair(result, mark.elapsedNow().inMilliseconds)
+    val elapsedTime = (System.nanoTime() - initialTime).inMilliseconds()
```
arnaudgiuliani, line 45: just simple extract as `startTime` & `endTime` those variables 👍

ppjeziellago: Done! 🚀

PR #889, opened 2020-08-30
Injection params added to dependency graph

arnaudgiuliani:
Add now the capacity to add parameters in the dependency graph:

allow to write from `factory { (id: String) -> SimpleViewModel(id, get()) }`
to `factory { SimpleViewModel(get(), get()) }`

both writings are available.

This will help inject silently API parameters (i.e: BundleState, WorkManager params ...) & allow compact writing to work:

`factory { (id: String) -> SimpleViewModel(id, get()) }`
can be written `factory<SimpleViewModel>()`


arnaudgiuliani:
not as easy ... it breaks the checkModules with default mocked values :/

Issue #892, opened 2020-09-03
How should I create a variable that allows null？


kdsrsl:
I have a variable defined as follows：
```
var answerTarget: IAnswerTarget? = null
```
use:
```
answerTarget?.start()
```
When answerTarget is null, start() is not executed. In line with expectations✅. 

-----

But when I use koin,
```
val answerTarget: IAnswerTarget? by inject()
```
use:
```
answerTarget?.start()
```

when I defined module
```
single<IAnswerTarget> { AnswerTargetImpl() }
```
start() is ok.

when i don't define module：

start() crash, throw exception:
```
org.koin.core.error.NoBeanDefFoundException: No definition found for class:'xxx.IAnswerTarget'. Check your definitions!
```
----

At this point, I want to allow answerTarget to be null without crashing.

Does koin support this?

Thx.

madroidmaq:
You can use the following methods:

```kotlin
val answerTarget: IAnswerTarget? by getKoin().injectOrNull()
```

At the same time, in order to simplify this process, I submitted a PR(#902) for more convenient use:

```kotlin
val answerTarget: IAnswerTarget? by injectOrNull()
```

senRsl:
> You can use the following methods:
> 
> ```kotlin
> val answerTarget: IAnswerTarget? by getKoin().injectOrNull()
> ```
> 
> At the same time, in order to simplify this process, I submitted a PR(#902) for more convenient use:
> 
> ```kotlin
> val answerTarget: IAnswerTarget? by injectOrNull()
> ```

This doesn't solve my problem.

PR #902, opened 2020-09-16
feat |> add getOrNull and injectOrNull in KoinComponent

madroidmaq:
During project development, some classes may not be injected but the program can still run normally without throwing `NoBeanDefFoundException` (which is exactly what the koin.getOrNull method does). I added an extension function similar to getOrNull to koiComponent in my own project. Later, when I read the project source code, I found that there are functions like `getOrNull` and `injectOrNull` in koin.
So I want to expose these two functions in koiComponent, so that people with the same needs as me will know the existence of these two functions, and can use them easily.

I’m not sure why these two methods were not exposed at the beginning. If this does not meet some design principles, please tell me why。


madroidmaq:
I found issues(#892) with the same demands as me.

koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinComponent.kt
```diff
@@ -54,6 +67,19 @@ inline fun <reified T> KoinComponent.inject(
 ): Lazy<T> =
     lazy(LazyThreadSafetyMode.NONE) { getKoin().get<T>(qualifier, parameters) }
 
+/**
+ * Lazy inject a Koin instance if available
+ * @param qualifier
+ * @param parameters
+ *
+ * @return Lazy instance of type T or null
+ */
```
arnaudgiuliani, line 76: Need to add @KoinApiExtension

koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinComponent.kt
```diff
@@ -43,6 +43,19 @@ inline fun <reified T> KoinComponent.get(
...
     getKoin().get(qualifier, parameters)
 
+/**
+ * Get a Koin instance if available
+ * @param qualifier
+ * @param parameters
+ *
+ * @return instance of type T or null
+ */
+inline fun <reified T> KoinComponent.getOrNull(
```
arnaudgiuliani, line 53: Need to add @KoinApiExtension

arnaudgiuliani (changes requested):
We are using @KoinApiExtension on KoinComponent. Please add it to those extra functions.

koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinComponent.kt
madroidmaq: Since it was developed based on the master branch before, I didn't notice this part. Now it has been corrected, thanks.

koin-projects/koin-core/src/main/kotlin/org/koin/core/KoinComponent.kt
madroidmaq: Since it was developed based on the master branch before, I didn't notice this part. Now it has been corrected, thanks.

madroidmaq:
> We are using @KoinApiExtension on KoinComponent. Please add it to those extra functions.

Since it was developed based on the master branch before, I didn't notice this part. Now it has been corrected, thanks.

rahul27:
@arnaudgiuliani wondering why this was closed?

Would be interested in an extension function like this.

Issue #903, opened 2020-09-18
koin-test-core module resolution fails 2.2.0-beta-1


niom:
I was just testing the latest 2.2.0-beta-1 release with one of my projects using Koin and found out that there's a problem in koin-test-core resolution when using at least then koin-test-junit5 and I expect that there's a similar problem with the koin-test.

Why is this happening is that I refactored the koin-test-core to be a separate module that contains the common classes and functionality and using implementing keyword for setting up the koin-test-core dependency to the koin-test and koin-test-junit5 modules. 

The correct keyword used should have been api instead of implementing. Totally my fault due i did not know well enough how transitive dependency handling works in gradle.

I write a patch for this during the week end.

For those who are interested gradle compile keyword was splitted in to two keywords: implementing and api.

using implementing keyword works as far as I understand like this:
``` 
koin-test-junit5
- implementing koin-test-core (private transitive)

and using this in the project
my-project
- implementing koin-test-junit5
-- implementing koin-test-core (only accessible by koin-test-junit5)
```

using the api keyword works like this:

``` 
koin-test-junit5
- api koin-test-core (public transitive)

and using this in the project
my-project
- implementing koin-test-junit5
-- api koin-test-core (accessible also from my-project)
```

niom:
I decided to use the same keyword as in all of the other koin project (compile) even though they are deprecated by gradle.

We want to update all the gradle dependency keywords at once.

niom:
Fixed

PR #907, opened 2020-09-21
Use same gradle dependency keywords

niom:
This fixes: https://github.com/InsertKoinIO/koin/issues/903

Use same gradle dependency inclusion keywords as the rest of the project.

implemeting -> compile
testImplementing -> testCompile

We should make a separate PR on converting all of the deprecated keywords to implementing, api, testImplementing in the future.


arnaudgiuliani:
Yes, seen that 👍 

thanks

PR #909, opened 2020-09-22
koin-androidx-workmanager

arnaudgiuliani:
Add integration of WorkManager, following the work of Fabio @fmatosqg

API is simplified. Also as WorkManager setup declaration.

Let's keep it minimalistic & see how it goes.


koin-projects/koin-androidx-workmanager/src/test/java/org/koin/androidx/workmanager/Classes.kt
```diff
@@ -0,0 +1,36 @@
+//package org.koin.androidx.workmanager
```
jeziellago, line 1: This file could be removed.

jeziellago:
All test code is fully comments. 👀

arnaudgiuliani:
> All test code is fully comments. 👀

yes, there is no need to test the factory. It's just a simple case for Koin here.

jeziellago:
> > All test code is fully comments. 👀
> 
> yes, there is no need to test the factory. It's just a simple case for Koin here.

Sure! 🚀

PR #911, opened 2020-09-23
Jetpack compose

arnaudgiuliani:
Add Koin integration for Jetpack Compose 🎉

Bring classical Android extensions to Compose composable functions: `get()`, `by inject()`, `getViewModel()` & `by viewModel()`


objcode:
Nifty.

Repeating convo from chat here – those functions won't be callable from a `@Preview` today context so the pattern of using them in composables that don't display UI seems spot on for now.

Example,

https://github.com/InsertKoinIO/koin/pull/911/files#diff-01d7a445d1abd94f7864f97c46e77521R36

arnaudgiuliani:
Need to add some docs :)

PR #914, opened 2020-10-08
Fix concurrency issue for Single definitions

Sloy:
# The bug
We were having some crashes in our production app when injecting a Segment SDK (analytics tracker). This SDK has the particularity that only one instance can be created with the same tag. ([Source for reference](https://github.com/segmentio/analytics-android/blob/master/analytics/src/main/java/com/segment/analytics/Analytics.java#L1376))

We are constructing this SDK inside a Koin's `single` definition. In theory, `single` only creates one instance of the bean, even in multi-threaded environments. But in our millions of daily sessions we get a few hundred crashes where this assumption fails.

We managed to reproduce the issue consistently in a unit test by simulating concurrent injections a few thousands of times. 🤯 
I'm submitting in the PR the most simple version of the test that I could write (our initial test was way more complicated 😂 ). The number of repetitions and threads come from trial and error, with lower numbers the test would sometime give a false positive. I'm not sure if you want to keep the test, or if this is the right place to put it. Let me know.

Screenshot 2020-10-08 at 10 49 17

# The cause
After some research, I came to the conclusion that the issue lies in the concurrency code from `SingleInstanceFactory`. 
The synchronization code in the `create()` function works fine, but there's one little operation that happens outside the `synchronized` block: the `value =` assignment.
The bug is so hard to reproduce because the affected code is such a tiny part! 
Screenshot 2020-10-08 at 11 17 24

# The fix
I surrounded the code in `get()` with another `synchronized`, and that seems to fix the issue. 🎉 
I'm no concurrency expert myself, but I can't think of any problems with this solution.

### DIFF
--- a/koin-projects/koin-core/src/main/kotlin/org/koin/core/instance/SingleInstanceFactory.kt
+++ b/koin-projects/koin-core/src/main/kotlin/org/koin/core/instance/SingleInstanceFactory.kt
@@ -44,8 +44,10 @@ class SingleInstanceFactory<T>(koin: Koin, beanDefinition: BeanDefinition<T>) :
 
     @Suppress("UNCHECKED_CAST")
     override fun get(context: InstanceContext): T {
-        if (!isCreated()) {
-            value = create(context)
+        synchronized(this) {
+            if (!isCreated()) {
+                value = create(context)
+            }
         }
         return value ?: error("Single instance created couldn't return value")
     }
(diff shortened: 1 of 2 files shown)


arnaudgiuliani:
Good catch. 

Then we don't need to keep it on create:

`
 override fun create(context: InstanceContext): T {
        return if (value == null) {
            super.create(context)
        } else value ?: error("Single instance created couldn't return value")
    }

    @Suppress("UNCHECKED_CAST")
    override fun get(context: InstanceContext): T {
        return synchronized(this) {
            if (!isCreated()) {
                value = create(context)
            }
            value ?: error("Single instance created couldn't return value")
        }
    }
`

arnaudgiuliani (changes requested):
Remoe sycnhonized from create 👍

Sloy:
Done!

arnaudgiuliani:
Great work 👍

arnaudgiuliani (approved):

PR #925, opened 2020-10-18
Add koin BOM java platform

tynn:
This PR adds a `koin-bom` module containing all other (release) modules.

The `publish-bintray.gradle` simplifies all other publishing scripts and enables the publication of the java platform. The `vcsUrl` now points to https://github.com/InsertKoinIO/koin

The absolute path to the gradle scripts using `$rootDir/gradle` has advantages over the relative path `../gradle`, as it can be used anywhere (even in another script).


arnaudgiuliani:
Really sorry to break your work @tynn . Now that we have updated everything for the new module architecture on master, and maven central migration you can send another proposal 🙏

tynn:
This is kind of obsolete now. I'll update this to base on the new implementation targeting maven central. I think a BOM will be nice anyhow.

arnaudgiuliani:
Good :)

helpermethod:
Hi @tynn, where can I find `koin-bom`? At least on Maven Central it doesn't seem to be available.

arnaudgiuliani:
this proposal is diverging too much :/ Let's open a new one if you want.

tynn:
@helpermethod I actually don't know how to handle this versioning scheme to create a BOM. Maybe @arnaudgiuliani has an idea. For me it looks like you'd have to provide a BOM for every single build `core`, `android`, etc. as well as one for the whole project. Otherwise you'd have to specify all the artifacts by name, which I don't really like.

arnaudgiuliani:
I would say that we don't need BOM: The idea is to provide all projects in the same major/minor version track. For example, you know all projects in 3.3 are compatible. 

Given version X.Y.Z, only Z can move to allow project to provide further updates patch

But else all projects are kind of linked together.

tynn:
@arnaudgiuliani the advantage of a BOM is that you don't need to think about that Z. You'd always have to keep track of the latest version in all projects. If you have a BOM, the latest version of each project would be used. The BOM is the link you're refering to.

arnaudgiuliani:
ok, good point then 👍 Let's rework on that

PR #926, opened 2020-10-19
Add GitHub Action's linter to check markdown files and remove invalid links

matir91:
Add linter to fix #922 

When a pr is opened the linter will run over all md files and check if a link is not valid.

# Notes:
- The linter will fail in the first commit because #922 is not merged (I want just to show how the errors are displayed)
- Some md files were removed because I couldn't file a redirection to them:
  - deleted:    coverpage.md
  - deleted:    developer-guides/android-developer-guide.md
  - deleted:    developer-guides/samples.md
  - deleted:    ../start/coverpage.md
- The removed link was removed from the doc, it doesn't make sense to have an invalid link.
For example:
  - https://www.ekito.fr/people/sparkjava-and-koin/
  - https://medium.com/mindorks/using-dependency-injection-with-koin-bee0b461714a
- Config Files comments:
  - CHANGELOG.md is excluded from the linter.
  It contains a lot of GitHub links, so when the linter checks it, GitHub returns max requests reached error (429 HTTP error code).
  The [linter library is adding a feature to fix this case. However, they have to release a new version with this feature](https://github.com/tcort/markdown-link-check/issues/114).
  - https://developer.android.com/ links are removed from the check because it requires authentication, so they fail due to a redirect loop.


matir91:
@arnaudgiuliani If we want to include this feature we should enable GitHub Actions

PR #962, opened 2020-11-26
Change _rootScope to val and remove _rootScopeDefinition

Jintin:
Get `_rootScope ` value from `_scope` and remove unused variable `_rootScopeDefinition`


Jintin:
Maybe the solution is not good enough, try new pr here:
https://github.com/InsertKoinIO/koin/pull/967

PR #967, opened 2020-11-28
Move create scope out of KoinApplication as it should inside ScopeRegistry

Jintin:
As title, we can move these two setup inside ScopeRegistry directly and make it private in order to form a better sctructure.

```
        koin.scopeRegistry.createRootScopeDefinition()
        koin.scopeRegistry.createRootScope()
```


Jintin:
Two follow up refactoring point: 
1. Move logger to things like GlobalContext so we don't have to access Koin object all the time.
2. Avoid pass Koin everywhere, child object is better not knowing the parent object if possible.

These two can be separate topic, let me know your comment if any.

arnaudgiuliani:
Your PR is out of sync. Please rework it against current master.

PR #968, opened 2020-11-29
Fix docs and sample codes for koin-androidx-workmanager

djyugg:
Fix two problems for koin-androidx-workmanager

- Wrong documentation
- An NoBeanDefFoundException that occurred on wrong worker definition in sample code


arnaudgiuliani:
as added in last 2.2.x version, injection parameters can be resolved directly in the graph with `get()`. Is it missing parameters?

djyugg:
> as added in last 2.2.x version, injection parameters can be resolved directly in the graph with get(). 

Sorry. I didn't know about that new feature.

But, in version 2.2.1 worker DSL can't resolve injection parameters with `get()`.
Maybe there is a problem with the worker DSL.

Revert changes related to the following issues
> An NoBeanDefFoundException that occurred on wrong worker definition in sample code

Also, the sample code in workmanager.md doesn't have enough arguments for MyListenableWorker.
Therefore, add an argument.

arnaudgiuliani:
Your PR pull too much commit, is either out of sync or based on previous work. Please try to resend something based on master

PR #999, opened 2021-02-01
Fix issue in Android - ViewModel Documentation

GMetaxakis:
In the current Documents the dependencies do not include the `org.koin:koin-android-viewmodel:$koin_version`.

In this PR we are changing this by including it along side with the `org.koin:koin-android:$koin_version` and we have replaced also the `compile` with `implementation`


arnaudgiuliani:
thanks

PR #1022, opened 2021-02-26
Setup publishing to Maven Central

cortinico:
This setups publishing for all the projects inside koin to Sonatype & `mavenCentral()`.

I've also cleaned up the old publishing as there were 3 different files for doing it :)

In order for this to work, you should setup the following property:
- `SONATYPE_NEXUS_USERNAME`: the sonatype username
- `SONATYPE_NEXUS_PASSWORD`: the sonatype password
- `GPG_SIGNING_KEY`: The ASCII-armored export of your private key (`gpg --export-secret-keys --armor <key_id>`)
- `GPG_SIGNING_PASSPHRASE`: The GPG passphrase.

If you're signing locally @arnaudgiuliani, you should rather let gradle use directly your GPG command and uncomment the `useGpgCmd` ([plus set the property required here](https://docs.gradle.org/current/userguide/signing_plugin.html#sec:using_gpg_agent))

I had a hard time testing this due to sonatype accesses. I believe there are still a couple of things to address:

- [ ] Update the group to `io.koin`
- [ ] Run a diff between the old and the new POM files
- [ ] Make sure that the Multiplatform projects are published correctly
- [ ] Make sure that the POM file satisfies Sonatype requirements.


arnaudgiuliani:
Thanks for you great contribution @cortinico 
I will make the tests 👍

martinbonnin:
For what it's worth, I've had issues with split staging repos publishing mpp projects, even when publishing from a single host. You might want to look into https://github.com/gradle-nexus/publish-plugin or using the nexus API directly to create an explicit staging repo ahead of publication.

cortinico:
> For what it's worth, I've had issues with split staging repos publishing mpp projects, even when publishing from a single host.

Hey @martinbonnin, which kind of issues are you talking about?

martinbonnin:
Mainly things like https://issues.sonatype.org/browse/MVNCENTRAL-940 and https://issues.sonatype.org/browse/OSSRH-63268 where the files are split across 2 staging repositories. That makes closing them impossible if for an example the jar is on one side and the signature on the other side.

I've been burned burned by this a bunch of times so maybe I'm overreacting there. Feel free to ignore my comment if everything is working fine!

cortinico:
Thanks for sharing 🙏 I haven't experienced anything like that but we'll keep an eye if that happens.

arnaudgiuliani:
Yep, let's keep an eye on it. Thanks @martinbonnin

PR #1032, opened 2021-03-03
Relax type resolution constraints

arnaudgiuliani:
Relax type resolution constraints.  Avoid forcing T:Any / KClass<T>

### DIFF
--- a/android-compose/koin-androidx-compose/build.gradle
+++ b/android-compose/koin-androidx-compose/build.gradle
@@ -53,7 +53,7 @@ dependencies {
 
     // Koin
     api "io.insert-koin:koin-android:$koin_version"
-    testImplementation "org.koin:koin-test-junit4:$koin_version"
+    testImplementation "io.insert-koin:koin-test-junit4:$koin_version"
     testImplementation 'junit:junit:4.13.1'
 
     /* Compose */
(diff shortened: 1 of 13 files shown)


AnonymousWalker:
@arnaudgiuliani 
In earlier version I could do this using KoinJavaComponent
`    val service = get(Service::class.java) `

However, since this change I have to add the explicit type because it no longer infers the generic type defined
`    val service: Service = get(Service::class.java) `

I am not sure why you change it that way

arnaudgiuliani:
what version do you talk about?

arnaudgiuliani:
this was effectively a problem in 2.2.x version, due to a "bad" patch

AnonymousWalker:
The latest 3.0.1-beta-2 changed to `clazz: Class<*>`
https://github.com/InsertKoinIO/koin/blob/master/core/koin-core/src/jvmMain/kotlin/org/koin/java/KoinJavaComponent.kt
where 3.0.1 was `clazz: Class<T>`
https://github.com/InsertKoinIO/koin/blob/3.0.1/core/koin-core/src/jvmMain/kotlin/org/koin/java/KoinJavaComponent.kt

AnonymousWalker:
I have an added extension that is available for global access, which was based on the extension methods in ktor pipeline:
[org.koin.ktor.ext.KoinApplicationCallExt](https://github.com/InsertKoinIO/koin/blob/master/ktor/koin-ktor/src/main/kotlin/org/koin/ktor/ext/KoinApplicationCallExt.kt)

Please find the proposed change here:
https://github.com/InsertKoinIO/koin/compare/master...AnonymousWalker:anonymouswalker-koin-common

PR #1036, opened 2021-03-04
Use qualified name of KClass to get class name in native

Yxovec:
Getting class name only by using simple name causes problems with duplicates when saving definitions into InstanceRegistry in multiplatform projects containing multiple classes with same simple name.


arnaudgiuliani (approved):
Ok, looks good 👍

PR #1041, opened 2021-03-05
bring back Koin gradle plugin

arnaudgiuliani:


arnaudgiuliani:
The plugin is not working for many Android cases. I keep it out for now

timrijckaert:
I don't like this decision to be honest. 

The plugin has been working just fine for our use-case....
Could you explain/link issues that were occurring; 

How are we supposed to know that dependencies are going to resolve at runtime?!
A DI/Service Locator library that "might" crash on you is a big deal breaker.
The fact that we could go incorporate a test that  catches these was the reason we went for Koin over Dagger/Hilt.

If issues arise with the plugin/test case it seems those need fixing instead of decommissioning the actual thing that catches errors.....

arnaudgiuliani:
if you have a bit of time to help continue on the topic, why not. Else I can just hold it like that, while we try something else. The solution is not covering enough. It's also very Android dedicated.

arnaudgiuliani:
I'm not closed to keep it like that for 3.0.1. There is a need to do something more.

PR #1042, opened 2021-03-05
make "bind" method type-safe

### LINKED ISSUE #1028: "bind" method is not type-safe
**Describe the bug**
`bind` extension method of `BeanDefinition` is not type-safe and can lead to runtime error, though it could be checked in compile-time in some cases.

**To Reproduce**
```kotlin
import org.koin.core.context.GlobalContext
import org.koin.core.context.startKoin
import org.koin.dsl.bind
import org.koin.dsl.module
import org.koin.test.check.checkModules

fun main() {
    val module = module {
        single { "the_string" }.bind<Int>()
    }

    checkModules {
        modules(module) // this won't help
    }

    startKoin {
        modules(module)
    }

    val koin = GlobalContext.get()

    println(koin.get<String>())
    println(koin.get<Int>()) // ClassCastException
}
```

**Expected behavior**
It would be great to move such errors to compile-time.

**Koin project used and used version:**
`koin_version = "2.2.2"`

### Possible solution
Type-safety may be achieved by using the following declaration, that specifies the base type explicitly:
```kotlin
infix fun <B : Any, T : B> BeanDefinition<T>.bindSafe(clazz: KClass<B>): BeanDefinition<T> {
    secondaryTypes = secondaryTypes + clazz
    return this
}
```
But there is a `bind` method, that doesn't take `KClass` argument and I didn't found yet an easy way to make it safe.
```kotlin
inline fun <reified T> BeanDefinition<*>.bind(): BeanDefinition<*> {
    return bind(T::class)
}
```
We can make the following:
```kotlin
inline fun <reified B : Any, T : B> BeanDefinition<T>.bindSafe(): BeanDefinition<T> {
    return bind(B::class)
}
```
But the usage of this in verbose:
```kotlin
single { "the_string" }.bindSafe<CharSequence, String>()
```
So this improvement is the subject for discussion.
I can make PR, if this solution will find support from community.

qwert2603:
Fixes #1028

### DIFF
--- a/core/koin-core/src/commonMain/kotlin/org/koin/core/Koin.kt
+++ b/core/koin-core/src/commonMain/kotlin/org/koin/core/Koin.kt
@@ -177,7 +177,7 @@ class Koin {
      *
      * @return instance of type S
      */
-    inline fun <reified S, reified P> bind(noinline parameters: ParametersDefinition? = null): S =
+    inline fun <reified S : Any, reified P : S> bind(noinline parameters: ParametersDefinition? = null): S =
             scopeRegistry.rootScope.bind<S, P>(parameters)
 
     /**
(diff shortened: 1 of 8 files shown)


qwert2603:
Maybe, method `inline fun <reified T> BeanDefinition<*>.bind(): BeanDefinition<*>` should be deprecated, as it is not type safe.

arnaudgiuliani:
Closing master branch closed all PRs :/ really sorry. Can you update your PR against main branch

qwert2603:
Hello! Main branch has gone far forward since this PR. So It will be easier to create another PR above main branch. I'll to do it a little bit later

PR #1050, opened 2021-03-19
Fix WorkManager docs

djyugg:
Fix wrong documentation for koin-androidx-workmanager

- Rewritten FragmentFactory to WorkManager


arnaudgiuliani:
Thanks

Issue #992, opened 2021-01-14
Workmanager library can't resolve WorkerParameters


fmatosqg:
**Describe the bug**
When the application tries to launch a particular worker in my closed source project the application crashes.

>  Caused by: org.koin.core.error.NoBeanDefFoundException: No definition found for class:'androidx.work.WorkerParameters'. Check your definitions!

Subsequent launches of the same app will crash immediately when the app opens, with same error message, even if we don't try to start any work. That's because the work previously scheduled is trying to run as soon as the application starts.

Depending on how the work is scheduled it will also crash the application repeatedly, even if it's not launched by user/debugger.

**To Reproduce**

Still under investigation.

One way to trigger the bug is changing the order of the constructor parameters on `ListenableWorker`. Then it happens at the moment the work is scheduled to start.

For example, in my project (closed source)

THIS worked 
**code A**:
```
class NewsDownloadWorker(
    context: Context,
    private val params: WorkerParameters,
    webService: WebService,
    private val settingsDao: RemoteTabDao,
    appDatabase: AppDatabase,
    workerTracker: WorkerTracker
)
```

while THIS failed

**code B**:
```
class NewsDownloadWorker(
    context: Context,
    webService: WebService,
    private val settingsDao: RemoteTabDao,
    appDatabase: AppDatabase,
    private val params: WorkerParameters,
    workerTracker: WorkerTracker
)
```

May be related to https://github.com/InsertKoinIO/koin/issues/988. Following the suggestions there, I found out that using named paramers I can work around the issue even when ''code B'' is used. This is the original dsl declaration:

**code 1:** 

```
  worker {
            NewsDownloadWorker(
               get(),
               get(),
                 get(),
                 get(),
               get(),
              get()
            )
        } 
```

this is the workaround I found:

**code 2:** 
```
  worker {
            NewsDownloadWorker(
                context = get(),
                params = get(),
                webService = get(),
                settingsDao = get(),
                appDatabase = get(),
                workerTracker = get()
            )
        } 
```

and this is the opposite of the workaround, meaning it will make it always fail:
**code 3**:
```
  worker {
            NewsDownloadWorker(
                context = get(),
                webService = get(),
                settingsDao = get(),
                appDatabase = get(),
                params = get(),
                workerTracker = get()
            )
        } 
```

In short:

- code A+1  works, but B+1 fails
- code A+2 and B+2 both work   (making code 2 a workaround)

[...]

PR #1055, opened 2021-03-28
Make `KoinWorkerFactory` match `androidx.work.WorkerFactory` API

fmatosqg:
Cherry-pick commit from master into 2.2.2   https://github.com/InsertKoinIO/koin/pull/1018 

> Origin `androidx.work.WorkerFactory` has `@Nullable` annotation on
abstract method `createWorker`. Inherited `KoinWorkerFactory` has
overriden return type of `createWorker` with non-null `ListenableWorker`.
In some edge cases that leads to
crash with `org.koin.core.error.NoBeanDefFoundException`.


arnaudgiuliani:
is it also the case in master branch?

fmatosqg:
Yep, same case as 3.x.x versions.

Do you think this may help with https://github.com/InsertKoinIO/koin/issues/992 not finding WorkParameters to be injected?

Issue #475, opened 2019-05-30
Register single by KClass argument


pauliusrum:
**Is your feature request related to a problem? Please describe.**

Hi,

I'm using koin in a ktor project. In that project we have data source classes (`UsersDataSource`, `PermissionsDataSource`). These classes are used for CRUD operations.
Frontend client makes a HTTP request in which it describes with which data source it wants to communicate. 
We use classgraph library to find all data source classes (they implement the same interface). Classgraph returns a `List<KClass<*>>` which we want to register as `singles`.
Currently there is no `single` method that takes `KClass` argument instead of `reified T` generic parameter. Also there should be `create` method that takes `KClass` instead of `reified T` generic parameter. 
These data source classes are also used in internal services so we want to be able to inject them using type parameters instead of `named`.

**Describe the solution you'd like**

The solution would be to add a `single` and `create` methods that take `KClass` (or java `Class` if interoperability is important) instead of `reified T` generic parameter and register the definition by the given `KClass`.

```Kotlin
package org.koin.core.module

fun <T : Any> single(
    type: KClass<T>,
    qualifier: Qualifier? = null,
    createdAtStart: Boolean = false,
    override: Boolean = false,
    definition: Definition<T>
): BeanDefinition<T> {
    // ...
}
```

```Kotlin
package org.koin.experimental.builder

inline fun <reified T : Any> create(context: Scope): T = create(T::class, context)

fun <T : Any> create(type: KClass<T>, context: Scope): T {
    lateinit var instance: T

    val ctor = type.getFirstJavaConstructor()
    val args = getArguments(ctor, context)
    instance = ctor.makeInstance(args)

    return instance
}
```

arnaudgiuliani:
Is available in koin-core-ext. You can do `single<MyClass>()`

pauliusrum:
https://github.com/InsertKoinIO/koin/tree/master/koin-projects/koin-core-ext/src/main/kotlin/org/koin/experimental/builder

There is only one function that takes `reified T`. I can not use it because the type is only known at runtime not compilation. We load these classes at runtime so we only have KClass<*>. Am I missing something?

arnaudgiuliani:
Ok I see. More a way to provide a single with KClass.

tynn:
Have you considered using the `Qualifier`, which is a parameter already, to achieve your goal?

From an implementation perspective it might be wise to have all DSLs delegate to a non-inline function to reduce inlined code and hide implementation details better.

[...]

PR #1093, opened 2021-05-14
KClass based instance builder function

ESchouten:
This PR enables KClass based instance creation, solving #475 
When working with application configuration files, dependency types are often not known at compile-time, which makes the use of generics not possible.
Instead, with this pull request, users can do as follows:
```
single {
    create(
        when (config.database) {
            Database.MEMORY -> InMemoryUserRepository::class
            Database.SQL -> SQLUserRepository::class
        }
    ) as UserRepository 
}
```


arnaudgiuliani:
Yep, usefull and already fixed in next 3.1.0. But taking it to deploy it on 3.0.2 👍

PR #1097, opened 2021-05-19
Update versions-android.gradle

Shusshu:
Update compose

### DIFF
--- a/gradle/versions-android.gradle
+++ b/gradle/versions-android.gradle
@@ -18,7 +18,7 @@ ext {
     
     androidx_workmanager = '2.5.0'
 
-    androidx_compose_gradle_version = "4.2.0-beta03" //"7.0.0-alpha08"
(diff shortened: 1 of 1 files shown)


Shusshu:
https://github.com/InsertKoinIO/koin/issues/1098

abhishekdewan101:
It would really nice to get this merged since without this we cant use compose beta-07 with koin

Shusshu:
Yes, that was a note in #1098 ... Does koin really need kotlin 1.5.0 that's the real question.

SergejIsbrecht:
Damn, looks like I did not see your PR. I opened up #1102 for exactly the same reason, but I just used `Kotlin` 1.4.32. I did a composite build with my project and it seems to work fine.

Shusshu:
Your PR is perfect
Closing this one as it is super-seeded by https://github.com/InsertKoinIO/koin/pull/1102

Issue #1092, opened 2021-05-13
Fragment scope problem of reuse 


ZuiRenA:
**Describe the bug**
When I use BottomNavigationView, two Fragments are used to display the page, and their ViewBinding both use Koin inject, but from Fragment A to Fragment B, and then from Fragment B to Fragment A, it throw `can' t get Scope for lifecycleOwner`

**My Code**
``` Kotlin
class MainActivity : BaseScopeActivity() {
....
   private val homeFragment: HomeFragment by inject()
   private val mineFragment: MineFragment by inject()

   fun  replaceCurrentPage(type: Int) {
       val targetFragment = if (type == 0) homeFragment else mineFragment
       supportFragmentManager.beginTransaction()
            .replace(R.id.mainContainer, targetFragment)
            .commit()
   }
....
}
```
``` Kotlin
class MineFragment : BaseScopeFragment<FragmentMineBinding>() {

    override fun initBinding(
        inflater: LayoutInflater,
        container: ViewGroup?
    ): Lazy<FragmentMineBinding> = inject { parametersOf(inflater, container) }
}
```

**solution**
When I debug fragmentScope, I found that _scope of LifecycleScopeDelegate will be set to null when the fragment is replaced. But because the _scope of LifecycleScopeDelegate will not be reassigned, an exception is thrown.

At first, I tried this, but I found that getValue is called faster than onCreate
``` Kotlin
@OnLifecycleEvent(Lifecycle.Event.ON_CREATE)
fun onCreate(owner: LifecycleOwner) {
     if (_scope != null && !_scope!!.closed) {
          return
     }
     val internalScopeId = owner.getScopeId()
     _scope = koin.getScopeOrNull(internalScopeId) ?: createScope(koin)
     logger.debug("Recreate scope: $_scope for $owner")
}
```

Then I can only use this way
```Kotlin
    override fun getValue(thisRef: LifecycleOwner, property: KProperty<*>): Scope {
//        return _scope ?: error("can't get Scope for $lifecycleOwner")
        if (_scope != null) return _scope!!

        val ownerState = thisRef.lifecycle.currentState
        val ownerIsActive =
            ownerState.isAtLeast(Lifecycle.State.CREATED)
        if (!ownerIsActive) {
            error("can't get Scope for $lifecycleOwner")
        }
        
        val koin = koinContext.get()
        _scope = koin.getScopeOrNull(thisRef.getScopeId()) ?: createScope(koin)
        return _scope!!
    }
```

[...]

PR #1101, opened 2021-05-24
Fix for reuse fragment scope

ZuiRenA:
#1092 I think it might be better for me to solve the reuse fragment in this way


arnaudgiuliani:
Let's try with your patch

Issue #1133, opened 2021-06-25
createdAtStart issue when the dependencies are overridden 


mirland:
**Describe the bug**
if you add the `createdAtStart` flag to one dependency, it's created at that moment. However, It should wait that all modules/depenencies are declared before invoke the create method

**To Reproduce**
```
interface Foo
class Foo1 : Foo
class Foo2 : Foo
class FooHolder(foo: Foo)


val module1 = module {
  single<Foo1>() bind Foo::class
}
val module2 = module {
  single(createdAtStart = true){
    val foo = get<Foo>()
    require(foo is Foo2) {
      "Foo should be an instance of Foo2"
    }
    FooHolder(foo)
  }
}
val module3 = module {
  single<Foo2>() bind Foo::class
}


startKoin {
   androidContext(this@App)
   modules(listOf(module1, module2, module3)) // It crashes because foo is instance of Foo1
}
```

**Expected behavior**
when `FooHolder` is injected, `foo` should be an instance of `Foo2` because it was overridden in the module3 

**Koin project used and used version (please complete the following information):**
koin-android version 3.1.1

Additional Info:
It may be related to https://github.com/InsertKoinIO/koin/issues/1131

mirland:
I close it for now because it's a special case of #1131, but we should keep it in mind

PR #1134, opened 2021-06-26
Fix how eager instances are created

### LINKED ISSUE #1131: NoBeanDefFoundException is thrown when createOnStart is used with dependencies of different modules
**Describe the bug**
If you have different dependencies declared in different modules, `Foo1` in `module1` and `Foo2` in `module2`.
Suppose that `Foo2` requires `Foo1` and also `Foo2` is created on start.
If `module2` is added before `module1`, the app crashes. 

Caused by: org.koin.core.error.NoBeanDefFoundException: No definition found for class 'class package.di.Foo1'

**To Reproduce**
```
class Foo1
class Foo2(val foo1: Foo1)

val module1 = module {
    single<Foo1>()
}


val module2 = module {
    single<Foo2>(createOnStart = true)
}
  
startKoin {
   androidContext(this@App)
   modules(listOf(module2, module1)) // It crashes!!!. However, modules(listOf(module1, module2))  works
}
```
It throws an exception because module2 was declared before module1

Caused by: org.koin.core.error.NoBeanDefFoundException: No definition found for class 'class package.di.Foo1'

**Expected behavior**
In my opinion, the order of the modules shouldn't be considered. This issue was introduced in the `3.X` version

**Koin project used and used version:**
 `koin-android version 3.1.1`

mirland:
Fixes #1131 and #1133

### DIFF
--- a/.github/workflows/github-actions-ci-cd.yml
+++ b/.github/workflows/github-actions-ci-cd.yml
@@ -1,5 +1,5 @@
 name: Koin CI/CD
-on: [push]
+on: [push, pull_request]
 jobs:
   Explore-GitHub-Actions:
     runs-on: ubuntu-latest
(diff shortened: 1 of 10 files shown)


mirland:
Fixed in https://github.com/InsertKoinIO/koin/commit/d3ed86624367a314109cf609724cb9ed1008c744

PR #1156, opened 2021-07-27
fix LocalSavedStateRegistryOwner.current is nullable

sonique6784:
LocalSavedStateRegistryOwner.current is nullable hence we need to check for nullability.
beside `T` might require to be changed to `T?` but that would mean many more changes


sonique6784:
not ready

PR #1166, opened 2021-08-11
Add BeanDefinition to ScopedInstance error message

Pwootage:
When creating an instance in a scoped context, if one of your dependencies is not found, it does not currently print out what it was trying to inject. This is a simple fix - just adds the BeanDefinition to the end of the error messages, like is done in one other spot in this file.


arnaudgiuliani:
thanks!

PR #1175, opened 2021-08-27
KoinJavaComponent.kt update to make it able to return null values.

peter-kovacs-accedo:
If the requested class is not available it will return with null.


ArbenMaloku:
@arnaudgiuliani Any plans to release this?

arnaudgiuliani:
Thanks!

Issue #1177, opened 2021-09-07
Koin compose is missing scopes support


dant3:
**Is your feature request related to a problem? Please describe.**
With version 3.1.2, given that we have a module:
```
module {
    scope<Authenticated> {
        viewModel<ProfileViewModel> { ProfileViewModelImpl(get()) }
    }
}
```
I can't get this viewModel in compose using koin-compose apis. The only function provided is 
```
inline fun <reified T : ViewModel> getViewModel(
    qualifier: Qualifier? = null,
    noinline parameters: ParametersDefinition? = null,
): T
```
Similar problem exists with get and inject methods.

**Describe the solution you'd like**
Ability to get viewModel within scope. Currently I had to develop custom API for that. The good side is that wasn't hard to do. What I ended up with is something like:
```

inline fun <reified T : ViewModel> ViewModelStoreOwner.getViewModel(
    scope: Scope?,
    qualifier: Qualifier? = null,
    noinline parameters: ParametersDefinition? = null,
): T {
    return getViewModel(scope, qualifier, T::class, parameters)
}

fun <T : ViewModel> ViewModelStoreOwner.getViewModel(
    scope: Scope? = null,
    qualifier: Qualifier? = null,
    clazz: KClass<T>,
    parameters: ParametersDefinition? = null,
): T {
    return when (scope) {
        null -> {
            GlobalContext.get().getViewModel(qualifier, null, { ViewModelOwner.from(this) }, clazz, parameters)
        }
        else -> {
            scope.getViewModel(qualifier, null, { ViewModelOwner.from(this) }, clazz, parameters)
        }
    }
}
```

**Target Koin project**
koin-copmpose

burnoo:
I've released a library - cokoin, where Koin Scopes are supported. It is an alternative to `org.koin:koin-androidx-compose`.

https://github.com/burnoo/cokoin

### [Library Migration Guide](https://github.com/burnoo/cokoin/wiki/Migration-from-org.koin:koin-androidx-compose)
### Use Cokoin with Scopes
Create or get scopes by wrapping composables with `@Composable KoinScope`:
#### API
```kotlin
@Composable
fun KoinScope(getScope: Koin.() -> Scope, content: @Composable () -> Unit) : Unit
```
#### Usage
```kotlin
object ScopeTypeA
object ScopeTypeB

data class A(val value: String)

val scopedModule = module {
    scope<ScopeTypeA> {
        scoped { A("scopeA") }
    }
[...]

Issue #1119, opened 2021-06-16
How to inject a viewModel into a Jetpack Compose Preview


FredericoNoce:
**Describe the bug**
I need to inject a Viewmodel into a composable function preview.

@Preview
    @Composable
    private fun PreviewHomeContent() {
        MyTheme {
            HomeContent(getViewModel())
        }
    }

java.lang.IllegalStateException: KoinApplication has not been started

@Preview
    @Composable
    private fun PreviewHomeContent() {
        **KoinComponent.initContext(LocalContext.current)**
        MyTheme {
            HomeContent(getViewModel())
        }
    }

**java.lang.IllegalStateException: Your activity is not yet attached to the Application instance. You can't request ViewModel before onCreate call.**

**Koin project used and used version (please complete the following information):**
org.koin:koin-android:2.2.0
org.koin:koin-androidx-viewmodel:2.2.0

headsvk:
Build your compose views in a way so they are testable without relying on view models.
See how Google does it https://developer.android.com/jetpack/compose/state#viewmodel-state
Then they can simply preview `HelloContent` with any parameters.

FredericoNoce:
Cool, I created a composable HomeContent method, in which it receives the viewmodel as a parameter. But he is injected by koin. With that, when I try to use the compose preview he has to "up" the context for him to be able to solve the injection, that would be my problem. @headsvk

burnoo:
I've released a library - cokoin, which supports Koin in `@Preview` . It is an alternative to `org.koin:koin-androidx-compose`.
### [Library Migration Guide](https://github.com/burnoo/cokoin/wiki/Migration-from-org.koin:koin-androidx-compose)
### Cokoin With `@Preview`
Wrap Preview Composable with `@Composable` Koin:
```kotlin
@Composable
fun Hello() {
     val hello = get<Hello>()
     Text(hello.text)
}

val previewModule = module {
     factory { Hello(text = "preview hello text") }
}

@Preview
@Composable
fun HelloPreview() {
    Koin(appDeclaration = { modules(previewModule) }) {
        Hello()
    }
}
```

Drjacky:
@headsvk I've created that extra `BlahContent` to use in the Preview but, the IDE didn't generate the UI.
P.S I use `hiltViewModel()`.

Issue #1072, opened 2021-04-24
Jetpack compose DefaultPreview fails


BulatMukhutdinov:
Using viewModel inside Composable funcions causes exception `java.lang.IllegalStateException: KoinApplication has not been started`.

I am injecting viewModel like this
```
@Composable
    private fun MyInfo(viewModel: MyInfoViewModel = getViewModel()) 
```

Koin version: 3.0.1

arnaudgiuliani:
Could be intersting to see ho we can avoid that. Can you specify a "dumb" ViewModel for preview?

jumaallan:
@arnaudgiuliani Any code samples? I am also facing the same issue

tamtom:
same issue 
Render problem 
`java.lang.IllegalStateException: KoinApplication has not been started   at org.koin.core.context.GlobalContext.get(GlobalContext.kt:36)   at org.koin.androidx.viewmodel.ext.android.ViewModelStoreOwnerExtKt.getViewModel(`

SirFilbido:
I'm having the same problem and the way I'm using it is in accordance with Koin's [documentation](https://insert-koin.io/docs/reference/koin-compose/compose).

```
@Composable
fun MyScreen(navController: NavController) {

    val viewModel = koinViewModel<MyViewModel>()

```

`java.lang.IllegalStateException: KoinApplication has not been started
`

Is there already a solution that doesn't use external libraries?

Issue #823, opened 2020-06-08
Calling inject() extension within @Composable functions  


henrikhorbovyi:
Playing around with Koin + Jetpack Compose, I would like to inject a dependency within a `@Composable` function.   

``` kotlin
@Composable
fun AppContent() {
    val newsRepository: NewsRepository by inject() // <~ it does not copile
    NewsScreen(newsRepository)
}
```
In this case, you get an error, because inject is an extension on `KoinComponent`, and there is no `KoinComponent` in this scope. 

So, I thought about a `@Composable` function that provides a `KoinComponent` scope.

 ``` kotlin 
object InjectableComponent : KoinComponent

@Composable
fun Injectable(child: @Composable InjectableComponent.() -> Unit) {
    child(InjectableComponent)
}
```
Using this `@Composable` would be like this:

``` kotlin
@Composable
fun AppContent() {
    Injectable { 
        val newsRepository: NewsRepository by inject()
        NewsScreen(newsRepository)
    }
}
``` 

I'd like to discuss this topic and hear opinions from you all

arnaudgiuliani:
If you use a standard Koin context, you can simply get Koin anywhere with `KoinContextHandler.get()`

henrikhorbovyi:
`KoinContextHandler` is in the `org.koin:koin-android`?

arnaudgiuliani:
in the main `koin-core` yes

henrikhorbovyi:
I'm using `core`

![image](https://user-images.githubusercontent.com/5937963/85034698-b4655680-b158-11ea-8e04-8ea1e485ceff.png)

But I can't find the `KoinContextHandler` :thinking:

arnaudgiuliani:
in 2.0.x, it's `KoinContext.get()`

henrikhorbovyi:
I don't know what I'm doing wrong, but I can't find KoinContext.get().

![image](https://user-images.githubusercontent.com/5937963/86145639-8d4e4380-bacd-11ea-811c-f86423fdeb98.png)

The most near that I get was: `GlobalContext`, is it the same? can I use it? :thinking:

henrikhorbovyi:
@arnaudgiuliani we've started a thread on [koin-dev](https://kotlinlang.slack.com/archives/CRF31N2HW/p1593531679064100). I'd like to hear your opinion :smile:

henrikhorbovyi:
[...]

PR #1178, opened 2021-09-13
[Proposal] New API for Jetpack Compose

burnoo:
I'm starting a discussion on `#koin-dev` - feel free to join  if you have any feedback or questions regarding this PR.

### Motivation
I’ve been working with Koin for Compose for a while now and have a proposal to improve how `KoinContext` management works in `@Composable` functions. I believe the configuration of the application should be done in the top level composable. This way it is not dependent on Koin initialization in the `Application` class, which allows for easy `KoinContext` replacement in `@Preview` composables and tests. It introduces a new API - `Koin` and `KoinScope` composables.

### Usage example (with scopes and @Preview):
```kotlin

object ScopeTypeA
object ScopeTypeB

data class A(val value: String)

val scopedModule = module {
    scope<ScopeTypeA> {
        scoped { A("scopeA") }
    }
    scope<ScopeTypeB> {
        scoped { A("scopeB") }
    }
}

@Preview
@Composable
fun Test() {
    Koin(appDeclaration = { modules(scopedModule) }) {
        Column {
            KoinScope(getScope = { createScope<ScopeTypeA>() }) {
                Text(get<A>().value)
            }
            KoinScope(getScope = { createScope<ScopeTypeB>() }) {
                Text(get<A>().value)
            }
        }
    }
}
```

### Pros:
- scopes works
- easy `KoinContext` replacement in tests
- easy `KoinContext` replacement in `@Preview` composables
- no need to use `Application` class (which may help with Compose Multiplatform support in the future)
- backward compatibility (using `GlobalContext` as default value)

### This PR includes:
- new API implementation
- instrumented tests
- necessary library updates
- fix of nullability issue in `ViewModelComposeExt.kt`

### Problems
- I couldn't update Compose to 1.0.2 because it uses newer version of Kotlin (1.5.21) than Koin
- to run the project I had to update `androidx_compose_gradle_version` to `7.0+`

### Issues fixed with this PR:
- https://github.com/InsertKoinIO/koin/issues/1177
- https://github.com/InsertKoinIO/koin/issues/1119
- https://github.com/InsertKoinIO/koin/issues/1072
- https://github.com/InsertKoinIO/koin/issues/823 (partially by `getKoinComponent().run { /* inject works here /* }`)

### TODO
- javadocs if proposal is accepted

### DIFF
--- a/android-compose/koin-androidx-compose/src/main/java/org/koin/androidx/compose/ComposeExt.kt
+++ b/android-compose/koin-androidx-compose/src/main/java/org/koin/androidx/compose/ComposeExt.kt
@@ -17,30 +17,33 @@ package org.koin.androidx.compose
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
-import org.koin.core.Koin
-import org.koin.core.context.GlobalContext
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.KoinScopeComponent
+import org.koin.core.component.get
 import org.koin.core.parameter.ParametersDefinition
 import org.koin.core.qualifier.Qualifier
 
-/**
- * Resolve a dependency for [Composable] functions
- * @param qualifier
- * @param parameters
- *
- * @return Lazy instance of type T
- *
- * @author Arnaud Giuliani
- * @author Henrique Horbovyi
- */
 @Composable
-inline fun <reified T> get(
-        qualifier: Qualifier? = null,
-        noinline parameters: ParametersDefinition? = null,
-): T = remember(qualifier,parameters) {
-    GlobalContext.get().get(qualifier, parameters)
+fun getKoinComponent(): KoinComponent {
+    val koin = getKoin()
+    val currentScope = getScope()
+    return currentScope?.let { scope ->
+        object : KoinScopeComponent {
+            override val scope = scope
+            override fun getKoin() = koin
+        }
+    } ?: object : KoinComponent {
+        override fun getKoin() = koin
+    }
 }
 
 @Composable
-fun getKoin(): Koin = remember {
-    GlobalContext.get()
+inline fun <reified T> get(
+    qualifier: Qualifier? = null,
+    noinline parameters: ParametersDefinition? = null,
+): T {
+    val scopeComponent = getKoinComponent()
+    return remember(qualifier, parameters) {
+        scopeComponent.get(qualifier, parameters)
+    }
 }
--- a/android-compose/gradle/wrapper/gradle-wrapper.properties
+++ b/android-compose/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.0-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.2-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
--- a/android-compose/koin-androidx-compose/build.gradle
+++ b/android-compose/koin-androidx-compose/build.gradle
@@ -13,8 +13,8 @@ ext {
 }
 
 android {
-    compileSdkVersion 29
-    buildToolsVersion "29.0.3"
+    compileSdkVersion 30
+    buildToolsVersion "30.0.3"
 
     defaultConfig {
         minSdkVersion 21
@@ -40,6 +40,7 @@ android {
         kotlinCompilerVersion kotlin_version
         kotlinCompilerExtensionVersion androidx_compose_version
     }
+
 }
 
 tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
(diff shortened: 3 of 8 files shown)


burnoo:
I've released this changes as standalone library for Compose Multiplatform. You can check it out here: https://github.com/burnoo/cokoin

arnaudgiuliani:
Yep, really appreciate your contribution. Could we add an update to https://github.com/InsertKoinIO/compose-samples?
It's a fork of base sample projects. It would help illustrate how we can play with Koin context in @Composable functions.

I can provide some guidelines for the javadoc.

android-compose/koin-androidx-compose/src/main/java/org/koin/androidx/compose/ComposeExt.kt
```diff
@@ -17,30 +17,33 @@ package org.koin.androidx.compose
...
- * @param qualifier
- * @param parameters
- *
- * @return Lazy instance of type T
- *
- * @author Arnaud Giuliani
- * @author Henrique Horbovyi
- */
 @Composable
-inline fun <reified T> get(
```
arnaudgiuliani, line 36: Why replace the `get` function? Either we can deprecate it or keep it 🤔

burnoo:
After opening this pull request, I thought that I want to do more with Koin for Compose. My first priority was to add Compose Multiplatform support. I also thought that `ViewModel` should be in the another artifact. I looked into issues related to Compose and found that Navigation is not supported.

With that in mind I've created my own library Cokoin - https://github.com/burnoo/cokoin. At first it was Multiplatform Compose, but then I've added Jetpack support. It contains the following features:
- this PR (Koin Composable + Scopes Support)
- Compose Multiplatform support
- separate artifact for `ViewModel`
- support for Navigation Compose (with separate artifact)

I think developing Koin for Compose as a separate library would have many benefits, such as:
- independence from the Koin version release cycle
- many artifacts would not overload Koin
- there would be no compatibility problems related to the Kotlin version
- easier keeping it up to date with the Jetpack Compose version
- Compose Multiplatform alpha support

I think that Cokoin could be official way of using Koin with Compose. I can do the following:
1. Add Javadocs to Cokoin
2. Create documentation page (docusaurus same as Koin)
3. Give you an opportunity to check the code (CR) and make adjustments if necessary
4. Add backward compatibility (for now in cokoin there is no backward compatibility, because `GlobalContext` is not in core, should be easy to add)
5. Update all Compose examples with Cokoin (and maybe move it to Cokoin repository)
6. (Maybe) Close this PR and deprecate `org.koin:koin-androidx-compose`

@arnaudgiuliani What do you think?

android-compose/koin-androidx-compose/src/main/java/org/koin/androidx/compose/ComposeExt.kt
burnoo: I only changed its place in code and removed the docs (by accident). It's here https://github.com/InsertKoinIO/koin/pull/1178/files/6062bdfdb3b2fc25f38acb26a0b8995b9480e508#diff-f9db5f6b8a1cbf8a2b9601a97856b228a48df598a9fb60c8e451a43b8ce5ccd2R41

ZeDzislaw:
On one side im happy that someone picked up compose koin DI problems, but on other side i'm worried that actual split from official repo will end up dead. Wouldn't be better to make effort to prepare PR to merge it to this repository? I understand some of your points for separate lib, but again - Im afraid of comunity dispersion.

burnoo:
@ZeDzislaw
Unfortunately this repository is not actively developed lately. Last Koin version was released 3 months ago. Compose is a very new technology and I think koin-compose releases should be much more frequent.

In my library I Introduced 5 more artifacts (and I am considering adding more for `SavedState` support). Adding them to this repository might not be the best idea.

I think a good compromise would be to move `cokoin` to InsertKoinIO GitHub, change artifacts name to `io.insert-koin:cokoin-*` (or other name) and put me in charge of managing releases. By doing so:
- I colud release the library frequently
- I could introduce a few artifacts and do not overload Koin
- cokoin would be a part of official GitHub, so would have community support

ZeDzislaw:
@burnoo I agree. I myself was concerned about the lack of new commits, especially in the context of jetpack compose. I appreciate your work on this subject and we will probably give some feedback in a couple of weeks as we plan to base the new screens on jetpack compose navigation in conjunction with koin.

apkelly:
Is the plan to move all of my Koin initialisation to the main Composeable in my Android app, as I don't think this covers some of the use-cases we need. Initialising my Koin modules in the Application class means I can inject my implementations into components that aren't Composables e.g. Services

burnoo:
@arnaudgiuliani Have you read my previous comments? What do you think?

@apkelly You are right. To use Koin in Services you have to initialize it in Application class. Sharing code between Service and Composables can be done by using same modules in Application and Koin Composable:
```kotlin
// Application class
override fun onCreate() {
    super.onCreate()
    startKoin { modules(sharedModule, servicesModule) }
}
```
```kotlin
@Composable
fun App() {
    Koin(appDeclaration = { modules(sharedModule, uiModule) }) {
       //...
    }
}
```

arnaudgiuliani:
> @arnaudgiuliani Have you read my previous comments? What do you think?

we can perhaps consider decoupling compose related development to help go faster. I'm open to help integrate in a better way

arnaudgiuliani:

PR #1186, opened 2021-10-16
Add `resolve` API to auto resolve constructors with no reflection

### LINKED ISSUE #1185: Auto resolve constructors with no reflection
I started to work last week as a consultant on a project which is 2 years old and was built with Koin. The project is relatively big and has tons of bindings, modules, and `get` method. I believe the current API to resolve dependencies is not very maintainable and has a lor of boilerplate. Any constructor change means we also need to change a binding. We can do better, auto resolving an instance by leveraging Kotlin's compiler.

By combining `reified` with constructor reference we can come up with a solution for the `get` without using reflection (which isn't a good option for Android). I would even argue that it makes completely redundant the reflection solution.

That is how it looks like today:

```kotlin
val myModule = module {
    viewModel {
        MyViewModel(
            dep1 = get(),
            dep2 = get(),
            dep3 = get(),
            dep4 = get()
        )
    }
```

And here is how it can look like:

```kotlin
val myModule = module {
    viewModel { resolve(::MyViewModel) }
}
```

The main advantages here are:
1. No need to boilerplate with `get()` functions.
2. No reflection, resolved at compile time.
3. Adding new parameters to a class does not require changing bindings.
4. Easy to find your bindings (search for `resolve(::YourDependency)`).
5. Compact, straightforward, and optional.

The code to make it work is very simple (I have been using it locally without issues):

```kotlin
inline fun <reified R, reified T1> Scope.resolve(
    constructor: (T1) -> R,
): R = constructor(get())

inline fun <reified R, reified T1, reified T2> Scope.resolve(
    constructor: (T1, T2) -> R,
): R = constructor(get(), get())

// other versions of it, up to 20 parameters.
```

Side-note: suggestions for the `resolve` function name are welcomed.

@arnaudgiuliani if you agree, I already submitted a PR with the changes - it is very straightforward.

Thank you.

marcellogalhardo:
Add constructor `resolve` functions, which let Koin take advantage of Kotlin's Compiler to deduce what parameters a class needs to resolve it at compile time without the need for reflection.

The API is as simple as it looks like, see the example below:

```kotlin
// Before
val myModule = module {
    viewModel {
        MyViewModel(
            dep1 = get(),
            dep2 = get(),
            dep3 = get(),
            dep4 = get()
        )
    }

// After
val myModule = module {
    viewModel { resolve(::MyViewModel) }
}
```

The main advantages here are:
1. No need to boilerplate with `get()` functions.
2. No reflection, resolved at compile time.
3. Adding new parameters to a class does not require changing bindings.
4. Easy to find your bindings (search for `resolve(::YourDependency)`).
5. Compact, straightforward, and optional.

Fixes #1185.

### DIFF
--- a/core/koin-core/src/commonMain/kotlin/org/koin/core/Resolver.kt
+++ b/core/koin-core/src/commonMain/kotlin/org/koin/core/Resolver.kt
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.koin.core.resolver
+package org.koin.core
 
 import org.koin.core.scope.Scope
 
--- a/core/koin-core/src/commonMain/kotlin/org/koin/core/resolver/Resolver.kt
+++ b/core/koin-core/src/commonMain/kotlin/org/koin/core/resolver/Resolver.kt
@@ -1,0 +1,171 @@
+/*
+ * Copyright 2017-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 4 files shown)


arnaudgiuliani:
Thank for your contribution 🙏 

As an extend let's chat about if we merge it into the main DSL

marcellogalhardo:
Right, in the next days I will create an issue so we can discuss next steps (integrating `new` on other bindings like `factory` and `single`).

Thank you @arnaudgiuliani! ✌️

PR #1198, opened 2021-11-01
fix package name of StateViewModelFactory

hoc081098:


arnaudgiuliani:
Nope, as it's helping us having visibility on right factory

hoc081098:
> Thanks for your proposal. Don't need any fix here, as it's helping us having visibility on right factory
Hi @arnaudgiuliani, what is "having visibility on right factory"?

arnaudgiuliani:
because we access android protected package, then we need to be at same package level

PR #1199, opened 2021-11-02
Refactors the String.clearQuotes() to use the String APIs instead of …

yuanchenxi95:
…manipulating the char array.


arnaudgiuliani:
thanks

PR #1206, opened 2021-11-04
Expand Resolve API to support `single`, `factory` and `viewModel`

marcellogalhardo:
Following up on the Pull Request of [constructor resolution](https://github.com/InsertKoinIO/koin/pull/1186) and @arnaudgiuliani [suggestion](https://github.com/InsertKoinIO/koin/pull/1186#issuecomment-959070131) as improvement, the following PR adds new methods for resolve `factory`, `single` and `viewModel` - including support for `qualifier` or not.

That is how it looks like:

```kotlin
module {
    singleOf(::UserRepository)
    factoryOf(::UserProperty, named("a"))
    viewModelOf(::UserViewModel)
}

// For comparison, before and after:
module {  
    viewModel(named("a")) { new(::MyViewModel) }
    viewModelOf(named("a"), ::MyViewModel)
}
```

For `single` and `factory` the solution is working fine, but for `viewModel` the previous syntax becomes unusable due to `Overload resolution ambiguity. All these functions match.`.

EDIT: The PR was updated to follow @arnaudgiuliani suggestion of naming by using the suffix `Of`. See the updated examples.


android/koin-android/src/main/java/org/koin/androidx/viewmodel/dsl/ModuleViewModelOfExt.kt
```diff
@@ -0,0 +1,244 @@
...
+ *   viewModel(::ViewModel)
+ * }
+ * ```
+ *
+ * @author Marcello Galhardo
+ *
+ * @see new
+ */
+inline fun <reified R : ViewModel> Module.viewModelOf(
+    qualifier: Qualifier,
```
marcellogalhardo, line 41: We can not use the default `null` parameter here, otherwise, the following syntax would not work: `viewModelOf(::MyViewModel)` as the `::MyViewModel` constructor would resolve on the `qualifier` and not on `constructor`.

arnaudgiuliani:
Thanks, @marcellogalhardo, yes `*Of` seems to be the right postFix. 

I see 2 things here:
 
- we should return directly the wrapped function to help reuse other DSL keywords, instead of just calling it. Like: 
`fun Module.factoryOf(...) = factory(...)`

this allow to write `factoryOf(::MyComponentImpl) bind MyComponent::class`

- concerning qualifier overload signature, I was thinking perhaps to have the constructor first and then nullable qualifier. 🤔

marcellogalhardo:
@arnaudgiuliani changes done for the parameters!

About the qualifier, I gave it a try and I find it very confusing that I can write `single(qualifier) { new(::Class) }` but with `singleOf` I need to write `singleOf(::Class, qualifier)`. The inconsistente feels weird.

Let me know if you want me to do the changes anyway. (:

marcellogalhardo:
@arnaudgiuliani all changes done.

arnaudgiuliani:
great work @marcellogalhardo  👍

arnaudgiuliani:
Just one thing to add: `scoped` definition, to help declare scoped definition. Can you add it?

marcellogalhardo:
> Just one thing to add: `scoped` definition, to help declare scoped definition. Can you add it?

Sure, I can do it.

marcellogalhardo:
> scoped

@arnaudgiuliani changes done.

arnaudgiuliani:

PR #1210, opened 2021-11-09
Implement Module Grouping

### LINKED ISSUE #1196: Module Grouping

When working on large projects, you normally have a nested package structure. Ideally, instead of having a single module that provides all dependencies, it would be nice to "compose" those modules into a single one.

Today you can do that by creating a list of modules, and connecting it in the "start Koin".

```kotlin
startKoin {
    modules(featureModule1 + featureModule2 + anotherModule)
}

val moduleCollection1 = listOf(featureDomainModule, featureDataModule, featurePresentationModule)

private featureDomainModule = module {}
private featureDataModule = module {}
private featurePresentationModule = module {}
```

The solution has some problems:
1. If you have multiple collections, you need to use the sum operator which will create a lot of unnecessary new lists behind the scenes.
2. If you have a collection of modules and singular modules, you need to concatenate those.
3. Add one by one in a larger project can easily become disorganized with developers choosing different approaches.

It would be nice to have built-in support from within the Koin library. It would look somehow like the following:

```kotlin
startKoin {
    modules(featureModule, featureModule2)
}

// Only exposed module, connected to `startKoin`
val featureModule = module {
    includes(featureDomainModule, featureDataModule, featurePresentationModule)
    // another name might be "dependsOn"
}

private featureDomainModule = module {}
private featureDataModule = module {}
private featurePresentationModule = module {}
```

The implementation idea here is that Koin would "automatically plug" all modules that are included when `modules` are called. That would favor more composition of modules and logical grouping by package.

Note: the solution would still create an internal list of "included modules", and we would need to recursively go down in the "included modules" as a module might include another module and so on...

What do you think?

marcellogalhardo:
Initial implementation for "[Module Grouping](https://github.com/InsertKoinIO/koin/issues/1196)" proposal.

It provides an API to include a `Module` as part of another `Module` in a DSL syntax. The implementation requires feedback (see [flatten](https://github.com/InsertKoinIO/koin/pull/1210/files#diff-e772f920a0a8f0cef44eb951fbb66027f8ee3b191a8cd91bc25b5f188ffaf837R342) method) to ensure we won't impact the load module in a negative way.

```kotlin
startKoin {
    modules(featureModule)
}

// Public `koin.Module` of the Gradle module.
public val featureModule = module {
    includes(featureDomainModule, featureDataModule, featurePresentationModule)
    // another name might be "dependsOn"
}

// Private `koin.Module` of the Gradle module.
// Internal details, should not be exposed.
private val featureDomainModule = module {}
private val featureDataModule = module {}
private val featurePresentationModule = module {}
```

Fixes #1196.

### DIFF
--- a/core/koin-core/src/commonMain/kotlin/org/koin/core/Koin.kt
+++ b/core/koin-core/src/commonMain/kotlin/org/koin/core/Koin.kt
@@ -324,14 +324,36 @@ class Koin {
      * Load module & create eager instances
      */
     fun loadModules(modules: List<Module>, allowOverride : Boolean = true) {
-        instanceRegistry.loadModules(modules, allowOverride)
-        scopeRegistry.loadScopes(modules)
+        val flattedModules = modules.flatten()
+        instanceRegistry.loadModules(flattedModules, allowOverride)
+        scopeRegistry.loadScopes(flattedModules)
         createEagerInstances()
     }
 
-
     fun unloadModules(modules: List<Module>) {
-        instanceRegistry.unloadModules(modules)
+        val flattedModules = modules.flatten()
+        instanceRegistry.unloadModules(flattedModules)
+    }
+
+    /**
+     * Returns a single list of all [Module] with their [includedModules] in the given list.
+     * Duplicated modules are ignored.
+     */
+    private fun List<Module>.flatten(): List<Module> {
+        val deque = ArrayDeque(elements = this)
+
+        val allModules = mutableListOf<Module>()
+        while (deque.isNotEmpty()) {
+            val module = deque.removeFirstOrNull()
+
+            // We want to ensure duplicated modules aren't visited more than once.
+            if (module != null && !allModules.contains(module)) {
+                allModules += module
+                deque += module.includedModules
+            }
+        }
+
+        return allModules
     }
 
     /**
--- a/android/koin-android/src/test/java/org/koin/test/android/IncludedModuleTest.kt
+++ b/android/koin-android/src/test/java/org/koin/test/android/IncludedModuleTest.kt
@@ -1,0 +1,45 @@
+package org.koin.test.android
+
+import org.junit.Test
+import org.koin.core.qualifier.named
+import org.koin.dsl.koinApplication
+import org.koin.dsl.module
+
+class IncludedModuleTest {
+
+    private class Person(parent: Person?)
+
+    private val QualifierA = named("a")
+    private val QualifierB = named("b")
+    private val QualifierC = named("c")
+    private val QualifierD = named("d")
+
+    private val ModuleA = module {
+        includes(ModuleB)
+        factory(QualifierA) { Person(parent = null) }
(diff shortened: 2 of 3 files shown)


core/koin-core/src/commonMain/kotlin/org/koin/core/Koin.kt
```diff
@@ -324,14 +324,37 @@ class Koin {
...
+        val flattedModules = modules.flatten()
+        instanceRegistry.unloadModules(flattedModules)
+    }
+
+    /**
+     * Returns a single list of all [Module] with their [includedModules] in the given list.
+     * Duplicated modules are ignored.
+     */
+    private fun List<Module>.flatten(): List<Module> {
+        val queue: Queue<Module> = LinkedList()
```
marcellogalhardo, line 343: Actually, we should use a `ArrayDeque` here. I will fix it tomorrow.

core/koin-core/src/commonMain/kotlin/org/koin/core/Koin.kt
```diff
@@ -324,14 +324,36 @@ class Koin {
...
+        val allModules = mutableListOf<Module>()
+        while (deque.isNotEmpty()) {
+            val module = deque.removeFirstOrNull()
+
+            // We want to ensure duplicated modules aren't visited more than once.
+            if (module != null && !allModules.contains(module)) {
+                allModules += module
+                deque += module.includedModules
+            }
+        }
```
arnaudgiuliani, line 354: Why not a `map` operation here? wouldn't be simpler to map a module and its included modules to a flat list?

Also, what about inner included modules?

arnaudgiuliani:
Hello @marcellogalhardo,

thanks for your work. We need to add unit tests to help be sure we cover the needed cases and we don't break it later.

core/koin-core/src/commonMain/kotlin/org/koin/core/Koin.kt
marcellogalhardo: I am not sure how a `map` would help here. Maybe you can expand a little bit there? 

Note that I'm considering the inner included modules (see below).

Basically, I'm considering each `module` in the list as a possible n-ary tree where each node can include sub-nodes. To iterate over it I'm doing a [Breadth-First Search](https://en.wikipedia.org/wiki/Breadth-first_search): for each node, I iterate over, I will include all `includedModules` to the `DequeArray` that will ensure all `includedModules` are in fact included. Any module that is already included won't be visited again. 

Maybe an image might be a little more helpful to understand what is going on here:

image

That is the best approach I know to do such operations in an n-ary tree like this one. If you are familiar with another approach, let me know and I will be happy to change it.

marcellogalhardo:
@arnaudgiuliani I included unit tests for the new feature in the file `IncludedModuleTest`. Let me know if you want me to add any specific case.

arnaudgiuliani:

PR #1226, opened 2021-11-19
Included modules

### LINKED ISSUE #1196: Module Grouping

marcellogalhardo:
Initial implementation for "[Module Grouping](https://github.com/InsertKoinIO/koin/issues/1196)" proposal.

It provides an API to include a `Module` as part of another `Module` in a DSL syntax. The implementation requires feedback (see [flatten](https://github.com/InsertKoinIO/koin/pull/1210/files#diff-e772f920a0a8f0cef44eb951fbb66027f8ee3b191a8cd91bc25b5f188ffaf837R342) method) to ensure we won't impact the load module in a negative way.

```kotlin
startKoin {
    modules(featureModule)
}

// Public `koin.Module` of the Gradle module.
public val featureModule = module {
    includes(featureDomainModule, featureDataModule, featurePresentationModule)
    // another name might be "dependsOn"
}

// Private `koin.Module` of the Gradle module.
// Internal details, should not be exposed.
private val featureDomainModule = module {}
private val featureDataModule = module {}
private val featurePresentationModule = module {}
```

Fixes #1196.
Original PR: https://github.com/InsertKoinIO/koin/pull/1210.

### DIFF
--- a/core/koin-core/src/commonMain/kotlin/org/koin/core/Koin.kt
+++ b/core/koin-core/src/commonMain/kotlin/org/koin/core/Koin.kt
@@ -324,13 +324,35 @@ class Koin {
      * Load module & create eager instances
      */
     fun loadModules(modules: List<Module>, allowOverride : Boolean = true) {
-        instanceRegistry.loadModules(modules, allowOverride)
-        scopeRegistry.loadScopes(modules)
+        val flattedModules = modules.flatten()
+        instanceRegistry.loadModules(flattedModules, allowOverride)
+        scopeRegistry.loadScopes(flattedModules)
     }
 
-
     fun unloadModules(modules: List<Module>) {
-        instanceRegistry.unloadModules(modules)
+        val flattedModules = modules.flatten()
+        instanceRegistry.unloadModules(flattedModules)
+    }
+
+    /**
+     * Returns a single list of all [Module] with their [includedModules] in the given list.
+     * Duplicated modules are ignored.
+     */
+    private fun List<Module>.flatten(): List<Module> {
+        val deque = ArrayDeque(elements = this)
+
+        val allModules = mutableListOf<Module>()
+        while (deque.isNotEmpty()) {
+            val module = deque.removeFirstOrNull()
+
+            // We want to ensure duplicated modules aren't visited more than once.
+            if (module != null && !allModules.contains(module)) {
+                allModules += module
+                deque += module.includedModules
+            }
+        }
+
+        return allModules
     }
 
     /**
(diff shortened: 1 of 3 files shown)


arnaudgiuliani:
there is conflict, because I pushed back the `createEagerInstances()` function

core/koin-core/src/commonMain/kotlin/org/koin/core/Koin.kt
```diff
@@ -324,13 +324,36 @@ class Koin {
      * Load module & create eager instances
      */
     fun loadModules(modules: List<Module>, allowOverride : Boolean = true) {
-        instanceRegistry.loadModules(modules, allowOverride)
-        scopeRegistry.loadScopes(modules)
+        val flattedModules = modules.flatten()
+        instanceRegistry.loadModules(flattedModules, allowOverride)
+        scopeRegistry.loadScopes(flattedModules)
+        createEagerInstances()
```
marcellogalhardo, line 330: @arnaudgiuliani can you double-check if that is correct? 🤔

marcellogalhardo:
@arnaudgiuliani changes done. Please, check it out and let me know if everything is good. Thank you! :v:

marcellogalhardo:
Hey @arnaudgiuliani, do you think we can merge this PR or should we do further modifications? 🤔

arnaudgiuliani:
Hi @marcellogalhardo just need to check performance impact here. I'm checking it

arnaudgiuliani:
let's go 👍

Issue #1240, opened 2021-12-03
ScopeDSL.viewModel is returning singleton instance of all view models


crocsandcoffee:
**Describe the bug**

ScopeDSL.viewModel is returning singleton instance of all view models 

See the code snippet below  for reference: 

```
inline fun <reified T : ViewModel> ScopeDSL.viewModel(
    qualifier: Qualifier? = null,
    noinline definition: Definition<T>
): Pair<Module, InstanceFactory<T>> {
    return scoped(qualifier, definition)
}
```

`return scoped(qualifier, definition)` is wrong and should be replace with `return factory(qualifier, definition)`

**Expected behavior**
A new instance of the view model should be returned instead of a singleton instance (unless this is a shared view model instance on Activity that is being injected into Fragment)

**Koin project used and used version (please complete the following information):**
v3.1.4

arnaudgiuliani:
Hello @crocsandcoffee 

it's not an error here, as the scoped VM instance is regenerated from the ViewModelFactory. It can then survive to Activity.

crocsandcoffee:
> Hello @crocsandcoffee 
> 
> it's not an error here, as the scoped VM instance is regenerated from the ViewModelFactory. It can then survive to Activity.

Here's what happens, I'm in Activity A and I launch Activity B, which has this scoped view model injected. I call Activity B finish() and return to Activity A. I start Activity B again, and the _same scoped view model instance_ is injected, it should be a new view model instance when a new activity is created. @arnaudgiuliani

thadcodes:
@arnaudgiuliani I created an example app that shows the issue here: https://gist.github.com/thadcodes/0c0cc293be80ddff38ebe76f12f68fe3

It allows you to specify a scope ID and text to show on the next screen.  You will notice that if you change the text field for a second call without changing the scope ID it returns the old text instead of one from a new view model even though I'm calling a new activity with new parameters and would expect it to create a new view model.  If you change the scope ID to create a new scope it then changes it.  This makes it impossible to use with a scope that spans multiple activities. 

I can provide example uses if you are still not convinced it's a bug

thadcodes:
> @arnaudgiuliani I created an example app that shows the issue here: https://gist.github.com/thadcodes/0c0cc293be80ddff38ebe76f12f68fe3
> 
[...]

PR #1252, opened 2021-12-16
ViewModel in scope to use Factory Declaration (#1)

thadcodes:
Fix for #1240


vitkhudenko:
Should the fix also affect Unit tests?

thadcodes:
Sorry I meant to close this after a slack conversation with @arnaudgiuliani because he is going to make this change.   But short answer is yes probably but I can't run the unit tests at all locally due to maven repo issues so I wasn't able to run the tests

vitkhudenko:
@thadcodes, thanks for explaining.

PR #1253, opened 2021-12-17
Replace LifecycleObserver with DefaultLifecycleObserver

octa-one:
Minor change. It's better to use DefaultLifecycleObserver instead of OnLifecycleEvent annotation, because LifecycleObserver with annotations uses reflection for method calls. And DefaultLifecycleObserver uses the default interface methods that work in Kotlin on any Java version.
And also in the latest version of androidx.lifecycle, OnLifecycleEvent annotation is considered deprecated https://developer.android.com/jetpack/androidx/releases/lifecycle#2.4.0


baraaezz (approved):
I see this commit useful as its more generic in terms of lifecycle

arnaudgiuliani (approved):

PR #1258, opened 2021-12-25
Kpg/kotlin nmm

kpgalligan:
Changes to support the new memory model for Kotlin/Native. I expect a period of transition between the two, so rather than simply change everything for the new memory model, the init code will preserve the old memory model design if the new memory model is not enabled. The new memory model will likely be the default later in 2022, at which point I'd just remove support for the old architecture.

For the strict memory model, the Koin global config was restricted to the main thread, to ensure that config remained mutable. The changes here for the new memory model make it act more like the JVM, which allows any thread access, and uses locks to safeguard concurrent access.

If it would be preferred to simply remove the old design, that will still work in the strict memory model, but it won't attempt to isolate everything to the main thread. That should work fine, unless the developer manages to freeze the global context. Still, for now I'd support both.

I do think we could consolidate the global context a bit, but maybe do that later when removing the strict model version.


kpgalligan:
Had to sign all commits. Posted a new pr.

PR #1261, opened 2021-12-25
Kpg/kotlin native memory model

kpgalligan:
Changes to support the new memory model for Kotlin/Native. I expect a period of transition between the two, so rather than simply change everything for the new memory model, the init code will preserve the old memory model design if the new memory model is not enabled. The new memory model will likely be the default later in 2022, at which point I'd just remove support for the old architecture.

For the strict memory model, the Koin global config was restricted to the main thread, to ensure that config remained mutable. The changes here for the new memory model make it act more like the JVM, which allows any thread access, and uses locks to safeguard concurrent access.

If it would be preferred to simply remove the old design, that will still work in the strict memory model, but it won't attempt to isolate everything to the main thread. That should work fine, unless the developer manages to freeze the global context. Still, for now I'd support both.

I do think we could consolidate the global context a bit, but maybe do that later when removing the strict model version.


core/koin-core/src/nativeMain/kotlin/org/koin/core/context/GlobalContext.kt
```diff
@@ -15,23 +15,26 @@
...
- * Global context - current Koin Application available globally
- *
- * Support to help inject automatically instances once KoinApp has been started
- *
- * @author Arnaud Giuliani
+ * Main thread only global context for the strict memory model
  */
-object GlobalContext : KoinContext {
-
+internal class StrictGlobalContxt: KoinContext {
```
xsveda, line 37:
```suggestion
internal class StrictGlobalContext: KoinContext {
```

core/koin-core/src/commonMain/kotlin/org/koin/mp/KoinPlatformTools.kt
```diff
@@ -12,6 +12,8 @@ expect object KoinPlatformTools {
...
     fun defaultLogger(level: Level = Level.INFO): Logger
     fun defaultContext(): KoinContext
-    fun <R> synchronized(lock: Any, block: () -> R): R
+    fun <R> synchronized(lock: SyncCapable, block: () -> R): R
     fun <K, V> safeHashMap(): MutableMap<K, V>
-}
\ No newline at end of file
+}
+
+expect open class SyncCapable()
```
arnaudgiuliani, line 19: I was thinking of `Lockable` instead of `SyncCapable` 🤔

arnaudgiuliani:
Hello @kpgalligan :) 

thank you very much for updating the KMM/Native side of Koin! 

I understand your implementation by providing Both new Koin context.

core/koin-core/src/commonMain/kotlin/org/koin/mp/KoinPlatformTools.kt
kpgalligan: Yeah, OK. I'll rename that.

arnaudgiuliani: thanks

arnaudgiuliani (approved):

arnaudgiuliani:
seems to have some conflicts 🤔

arnaudgiuliani:
Great job Kevin :)

PR #1264, opened 2022-01-01
Update the comment for `declare`

istvankovacs-priv:
### What?
The comment for the `declare` function was updated to indicate the actual default value of the `allowOverride` parameter.

### Why?
The default value was changed to `true` by this [PR](https://github.com/InsertKoinIO/koin/commit/c3fdd69612f239a9f5e5f2e16d1a1b75fc443139), but the comment was accidentally left intact.


arnaudgiuliani:
thanks

Issue #1257, opened 2021-12-23
Koin Plugin Does Not Work With Ktor 2.0


AlexandreBrown:
**Describe the bug**
Some breaking changes were introduced in ktor 2.0.  
Ktor "features" were renamed to "plugins" and this causes Koin Feature to be incompatible with Ktor 2.0.
More information about this change can be found [here](https://ktor.io/docs/eap/migrating-2.html#feature-plugin).

**To Reproduce**
1. Create a Ktor 2.0 project
2. Add Koin ktor dependency : `implementation("io.insert-koin:koin-ktor:3.1.4")`
3. Try to install the Koin Plugin  
```kotlin
import io.ktor.server.application.Application
import org.koin.ktor.ext.Koin

fun Application.installPlugins() {
	install(Koin)
}
```
4. See error, the Koin Plugin cannot be installed.  
![image](https://user-images.githubusercontent.com/26939775/147251034-2a897aec-5503-44b7-90a1-5da85deb739b.png)
```kotlin
Type mismatch.
    Required:
        Plugin<TypeVariable(P), TypeVariable(B), TypeVariable(F)>
    Found:
        Koin.Feature
```

**Expected behavior**
The Koin Plugin can be installed.

**Koin project used and used version:**
`koin-ktor version 3.1.4`

**Additional moduleDefinition**
`io.ktor:ktor-server-core:2.0.0-beta-1`

phucynwa:
I got the same problem. Please try this.

```kotlin
object KoinPlugin : ApplicationPlugin<Application, KoinApplication, Koin> {

    override val key: AttributeKey<Koin>
        get() = AttributeKey("Koin")

    override fun install(
        pipeline: Application,
        configure: KoinApplication.() -> Unit
    ): Koin {
        val monitor = pipeline.environment.monitor
        val koinApplication = startKoin(appDeclaration = configure)
        monitor.raise(KoinApplicationStarted, koinApplication)

        monitor.subscribe(ApplicationStopping) {
            monitor.raise(KoinApplicationStopPreparing, koinApplication)
            stopKoin()
            monitor.raise(KoinApplicationStopped, koinApplication)
        }

        return Koin(koinApplication)
    }
}
```

https://gist.github.com/phucynwa/bc29d8e5dc9b0f1eacd2dc6101dbb0b6

AlexandreBrown:
[...]

PR #1266, opened 2022-01-06
Add Ktor 2.0 Plugin for Koin

phucynwa:
Since version 2.0, Ktor uses Plugin instead of Feature.

Refs #1257

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -14,6 +14,9 @@ ext {
     // Ktor
     ktor_version = '1.6.4'
 
+    // Ktor 2.0
+    ktor2_version = '2.0.0-beta-1'
+
     // Test
     junit_version = "4.13.2"
     junit5_version = "5.6.2"
--- a/ktor/koin-ktor2/build.gradle
+++ b/ktor/koin-ktor2/build.gradle
@@ -1,6 +1,6 @@
 apply plugin: 'kotlin'
 
-archivesBaseName = 'koin-ktor'
+archivesBaseName = 'koin-ktor2'
 
 repositories {
     mavenCentral()
@@ -11,8 +11,7 @@ dependencies {
     testImplementation "io.insert-koin:koin-test-junit4:$koin_version"
 
     // Ktor
-    compile "io.ktor:ktor-server-core:$ktor_version"
-    testImplementation "io.ktor:ktor-server-test-host:$ktor_version"
+    compile "io.ktor:ktor-server-core:$ktor2_version"
 }
 
 apply from: '../../gradle/publish-to-central.gradle'
--- a/ktor/koin-ktor2/src/main/kotlin/org/koin/ktor2/ext/KoinApplicationEvents.kt
+++ b/ktor/koin-ktor2/src/main/kotlin/org/koin/ktor2/ext/KoinApplicationEvents.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.koin.ktor2.ext
+
+import io.ktor.events.*
+import org.koin.core.KoinApplication
+
+/**
+ * @author Phuc YNWA
+ */
+
+/**
+ * Event definition for [KoinApplication] Started event
+ */
+val KoinApplicationStarted = EventDefinition<KoinApplication>()
+
+/**
+ * Event definition for an event that is fired when the [KoinApplication] is going to stop
+ */
+val KoinApplicationStopPreparing = EventDefinition<KoinApplication>()
+
+/**
+ * Event definition for [KoinApplication] Stopping event
+ */
+val KoinApplicationStopped = EventDefinition<KoinApplication>()
--- a/ktor/koin-ktor2/src/main/kotlin/org/koin/ktor2/ext/KoinApplicationExt.kt
+++ b/ktor/koin-ktor2/src/main/kotlin/org/koin/ktor2/ext/KoinApplicationExt.kt
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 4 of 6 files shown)


arnaudgiuliani:
Seems still in beta, no? 🤔

phucynwa:
@arnaudgiuliani Ktor 2.0 is still in beta. However, many Ktor project using Koin as a DI framework had migrated to 2.0 and cannot use Koin anymore. We should support them. Maybe in the beta version of Koin 3.1.6 ?

AlexandreBrown:
@arnaudgiuliani FYI 
image
Source : Kotlin Slack

arnaudgiuliani:
Yeah, great thing. First beta version of Koin 3.2 is starting those days until start of April (beta phase of 6 weeks). We can add Ktor 2.0 in it 👍

CloudyDino:
Any chance this can be merged in now that Koin 3.2 beta-1 has been released?

alexfacciorusso:
Given this seems to be a non-destructive change (e.g. it's just adding the plugin), is there any reason why this PR doesn't get taken into consideration? Thanks!

arnaudgiuliani:
Need a last check on it, but seems ok 👍  Any impact on ktor sample app?

phucynwa:
@arnaudgiuliani No, I created a new module for Ktor 2. No affect to the exist samples.

arnaudgiuliani:
the interesting point here is to decide either to go Ktor 2.0 only, either to offer both. The new 3.2 branch is dedicated to new versions, that I would keep Ktor 1.6.x into Koin 3.1.x and go with ktor 2.0 in 3.2. I need to check with ktor team if their will maintain other branches of Ktor 1.x

alexfacciorusso:
> the interesting point here is to decide either to go Ktor 2.0 only, either to offer both. The new 3.2 branch is dedicated to new versions, that I would keep Ktor 1.6.x into Koin 3.1.x and go with ktor 2.0 in 3.2. I need to check with ktor team if their will maintain other branches of Ktor 1.x

I believe it would be about keeping back-compatibility: many backends would continue using Ktor 1.x still for long time (e.g. they won't update immediately to Ktor 2 since upgrading to it might require big efforts from their side, especially for production servers and big projects), and removing the support for Ktor 1.x would mean leaving many projects without first-citizen support from Koin, which I believe would be a huge loss.

If my personal opinion would count something, I would advise against removing the support for the 1.x and instead add the v2 as additional plugin (what this PR is doing). Absolutely feel free to liaise with the Ktor team, maybe they would have other useful input

arnaudgiuliani:
Ktor 1.x will be maintained for security patches. Those patches will be ported to Koin 3.1.x - https://medium.com/koin-developers/koin-3-2-beta-roadmap-for-2022-b3b4afea69d8

The problem now is: is there really a need to extract koin-ktor from the main release cycle, to have its own release cycle like compose or Koin annotations.

alexfacciorusso:
> Ktor 1.x will be maintained for security patches. Those patches will be ported to Koin 3.1.x - https://medium.com/koin-developers/koin-3-2-beta-roadmap-for-2022-b3b4afea69d8
> 
> The problem now is: is there really a need to extract koin-ktor from the main release cycle, to have its own release cycle like compose or Koin annotations.

I believe that would definitely benefit the Ktor community. That said, how would the Gradle/Maven dependency be named? 
`io.insert-koin:koin-ktor2` and `io.insert-koin:koin-ktor3`, where `ktor` would initially point to ktor 2 and then ktor 3 whenever it will be released?

If so, then the documentation would need to put an alert or something to let people know of the change

alexfacciorusso:
My suggestion is because if you simply stop updating the `io.insert-koin:koin-ktor` library, no one would know of the change of name, if it makes sense

arnaudgiuliani:
by extracting koin-ktor project?

alexfacciorusso:
> by extracting koin-ktor project?

yes. Extracting the ktor "plugin" in its own library it means the Ktor community can have a faster updates cycle if needed

arnaudgiuliani:
Yes, I will extract koin-ktor development in its own project for Koin 3.2 timeline. It will be better.

CloudyDino:
So if I'm on Ktor 2.0, will I get the update on the `io.insert-koin:koin-ktor` library or will there be a different library that I'll have to import? Just not sure what the result of extracting the koin-ktor library into its own project means for Ktor devs.

arnaudgiuliani:
it will be the same `io.insert-koin:koin-ktor`, from version 3.2 ok Koin. We will catchup ktor 2.0 cycle.

ktor 1.6 -> Koin 3.1
ktor 2.0 -> Koin 3.2

phucynwa:
Ktor 2.0.0 RC-1 has been released.

geek5nan:
Not only Ktor 2.0.0-rc1 has been released, but Ktor 2.0.0 has also been released silently. 

[2.0.0](https://repo1.maven.org/maven2/io/ktor/ktor/2.0.0/) - 2022-04-08 07:10         
[2.0.0-beta-1](https://repo1.maven.org/maven2/io/ktor/ktor/2.0.0-beta-1/)   -  2021-12-22 11:27
[2.0.0-rc-1](https://repo1.maven.org/maven2/io/ktor/ktor/2.0.0-rc-1/) - 2022-04-04 13:42

arnaudgiuliani:
from what I see, we don't need 2 packages of koin-ktor, but rather the same that evolves with Ktor versions.

phucynwa:
I think I'll update this PR soon.

arnaudgiuliani:
Let me create the new repo, and extract base project for it. Once ready we can move the PR

Marek00Malik:
What is the status of this PR? 
We have to work with hacks like this [one](https://github.com/InsertKoinIO/koin/issues/1295) to work on Ktor 2.0.0.

jmfayard:
@phucynwa could you update your PR with this gist ? https://gist.github.com/MarkusNowotny/c245b5f493bfdde09d7b9853ef91fb55

markiewiczart:
KoinPlugin in not compatible with Ktor 2.0.1 anymore as the interface `ApplicationPlugin` changed. The new way of creating a plugin would be something like this:

```kotlin
// Config class
class KoinPluginConfiguration {
    var appDeclaration: KoinAppDeclaration = {}
        private set

    fun koin(block: KoinAppDeclaration) {
        appDeclaration = block
    }
}

// Plugin
val KoinPlugin = createApplicationPlugin(name = "KoinPlugin", createConfiguration = ::KoinPluginConfiguration) {
    val monitor = environment?.monitor
    val koinApplication = startKoin(appDeclaration = pluginConfig.appDeclaration)
    monitor?.raise(KoinApplicationStarted, koinApplication)

    monitor?.subscribe(ApplicationStopping) {
        monitor.raise(KoinApplicationStopPreparing, koinApplication)
        stopKoin()
        monitor.raise(KoinApplicationStopped, koinApplication)
    }
}

// Usage
fun Application.configureKoin() {
    install(KoinPlugin) {
        koin {
            logger(SLF4JLogger())
            modules(
                myModule(),
                module {
                    single { MyService() }
                }
            )
        }
    }
}
```

Extension functions work fine.

arnaudgiuliani:
thanks, very helpful! I'm working on it 💪

arnaudgiuliani:
Project is now ready for Ktor 2 in https://github.com/InsertKoinIO/koin-ktor

PR #1271, opened 2022-01-13
Inline log function with message lambda

octa-one:
Minor improvement.
When calling `log(level: Level, msg: () -> MESSAGE)` function with an **inline** modifier, no lambda object is created. This will slightly improve performance.
I also removed canLog(...) function, because it is completely similar to the isAt(...) function.

### DIFF
--- a/core/koin-core/src/commonMain/kotlin/org/koin/core/logger/Logger.kt
+++ b/core/koin-core/src/commonMain/kotlin/org/koin/core/logger/Logger.kt
@@ -24,10 +24,8 @@ abstract class Logger(var level: Level = Level.INFO) {
 
     abstract fun log(level: Level, msg: MESSAGE)
 
-    private fun canLog(level: Level): Boolean = this.level <= level
-
     private fun doLog(level: Level, msg: MESSAGE) {
-        if (canLog(level)) {
+        if (isAt(level)) {
             log(level, msg)
         }
     }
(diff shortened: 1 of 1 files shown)


arnaudgiuliani:
what about perf impact? I'm always worried about impact as such level

octa-one:
Sorry for late reply. What do you mean about performance impact?
If we add an inline modifier, as I suggest, Kotlin compiler will place the condition `if (isAt(level)) { }` from the function body at the call site, and will not create an extra instance for the lambda in function arguments. 

Less object will be created, which will have a positive effect on performance, especially considering that logging is called quite often.

arnaudgiuliani:
Need to check in our latest version of Kotlin, but string evaluation could be done even if the string is in a lambda: like `log() { "a message here" }`. The message could be even evaluated. This is what we don't want, as the logs that aren't already inlined (because there's is already an API for that) mustn't evaluate messages that could provide heavier information load.

octa-one:
Didn’t change anything, forgot that I already have 2 branches with pull requests.

ANIKINKIRILL:
why this PR is still not merged? what does stop this?

arnaudgiuliani:
It's planned for `3.3.0` as part of new internal. Then, not released for `3.2.1` where we bring fixes only & library patch updates.

arnaudgiuliani:
Log API has been revamped https://github.com/InsertKoinIO/koin/blob/core/3.3.0/core/koin-core/src/commonMain/kotlin/org/koin/core/logger/Logger.kt

Thanks for your proposal. Feel free to reopen discussion on new API

PR #1286, opened 2022-02-04
Add navigation args to `SavedStateHandle`

rubenquadros:
Navigation arguments can be passed when navigating to a destination. These arguments can be passed down to `SavedStateHandle` which can be accessed directly in the `ViewModel`.

### DIFF
--- a/android/koin-android/build.gradle
+++ b/android/koin-android/build.gradle
@@ -22,6 +22,7 @@ android {
 
 dependencies {
     api "io.insert-koin:koin-core:$koin_version"
+    api "androidx.navigation:navigation-runtime-ktx:$androidx_nav"
 
     implementation "androidx.appcompat:appcompat:$androidx_lib_version"
 
--- a/android/koin-android/src/main/java/org/koin/androidx/viewmodel/factory/StateViewModelFactory.kt
+++ b/android/koin-android/src/main/java/org/koin/androidx/viewmodel/factory/StateViewModelFactory.kt
@@ -13,6 +13,11 @@ class StateViewModelFactory<T : ViewModel>(
     parameters.state?.invoke()
 ) {
     override fun <T : ViewModel> create(key: String, modelClass: Class<T>, handle: SavedStateHandle): T {
+        parameters.defaultArgs?.let { args ->
+            args.keySet().forEach {
+                handle[it] = args.get(it)
+            }
+        }
         val params: ParametersDefinition = addHandle(handle)
         return scope.get(
             parameters.clazz,
--- a/android/koin-android/src/main/java/org/koin/androidx/viewmodel/ViewModelOwner.kt
+++ b/android/koin-android/src/main/java/org/koin/androidx/viewmodel/ViewModelOwner.kt
@@ -1,7 +1,9 @@
 package org.koin.androidx.viewmodel
 
+import android.os.Bundle
 import androidx.lifecycle.ViewModelStore
 import androidx.lifecycle.ViewModelStoreOwner
+import androidx.navigation.NavBackStackEntry
 import androidx.savedstate.SavedStateRegistryOwner
 
 typealias ViewModelOwnerDefinition = () -> ViewModelOwner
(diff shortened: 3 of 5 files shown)


wiryadev:
Excuse me but i just want to remind that (AFAIK) your commit needs to be signed first before being able to merged

rubenquadros:
Really sorry, wasn't aware of this. I have pushed a signed commit now.

android/koin-android/build.gradle
```diff
@@ -22,6 +22,7 @@ android {
 
 dependencies {
     api "io.insert-koin:koin-core:$koin_version"
+    api "androidx.navigation:navigation-runtime-ktx:$androidx_nav"
```
arnaudgiuliani, line 25: navigation runtime can't be addded here. Another project is helping with it: koin-androidx-navigation.

arnaudgiuliani (changes requested):
Need to check with koin-androidx-navigation project if it's not already added. But here you can't add this dependency.

android/koin-android/build.gradle
rubenquadros: can we add navigation project dependency in compose or we create a new project with compose + navigation dependency?

arnaudgiuliani:
Your PR is in the checklist for 3.2. I'll ping you when in a few days

android/koin-android/src/main/java/org/koin/androidx/viewmodel/factory/StateViewModelFactory.kt
```diff
@@ -13,6 +13,11 @@ class StateViewModelFactory<T : ViewModel>(
     parameters.state?.invoke()
 ) {
     override fun <T : ViewModel> create(key: String, modelClass: Class<T>, handle: SavedStateHandle): T {
+        parameters.defaultArgs?.let { args ->
+            args.keySet().forEach {
+                handle[it] = args.get(it)
+            }
+        }
```
raamcosta, line 20: Why do we need to do this here? The usual way to do this is to pass this bundle in the `AbstractSavedStateViewModelFactory` constructor parameter.
I see that we are passing a bundle from `ViewModelParameter` state lambda.
So I wonder: why do these two intersect? When is the `parameters.state` lambda used vs the new `defaultArgs`? If there is a key present in both, then the one in `defaultArgs` will override the other, is that expected?

If they serve the same purpose, shouldn't they be one only field of `ViewModelParameter` instead of two?

Also, even if we want to do things like this, for some reason, I would do this:

```kotlin
) : AbstractSavedStateViewModelFactory(
    parameters.registryOwner ?: error("Can't create SavedStateViewModelFactory without a proper stateRegistryOwner"),
    // 👇 probably deserves its own function, and code is not tested but hopefuly it's enough to understand my point
    parameters.state?.invoke()?.let { state -> 
        if (parameters.defaultArgs != null) state.putAll(paramters.defaultArgs)
        else state
    } ?: parameters.defaultArgs
) {
```

yschimke:
I've just hit this and need this, what is the API needed to use this? Is it going to automagically work to pass in navigation arguments?

android/koin-android/src/main/java/org/koin/androidx/viewmodel/factory/StateViewModelFactory.kt
arnaudgiuliani: Koin injected parameters are not necessarily bundle args. But here the API become to be outdated, and perhaps we don't both side

rubenquadros: After looking at it again I don't think we would require the new `defaultArgs` here. We can make use of `parameters.state` which is being passed to the `AbstractSavedStateViewModelFactory` constructor.

rubenquadros: I have tried it in a [sample project](https://github.com/rubenquadros/Koin-Compose-Sample/commit/fda566a7f20c085c7eccbcb28e18e54118fb1804) (commit is a bit messy 😅) and it's working as expected. However, it is with the 3.1.4 version. Will try it with updated code.

arnaudgiuliani:
I've just updated the androidx-compose project to update everything, to follow with right `owner` and pass `parameters`

arnaudgiuliani:
It zill be deployed in next androidx-compose project

Gwindor:
@arnaudgiuliani What is the best way to get a fix for that issue? Having navigation arguments is pretty common.

PR #1287, opened 2022-02-09
Don't cast to AppCompatActivity to support FragmentScenario

DonJogi:
Using ComponentActivity instead of AppCompatActivity in the
LifecycleViewModelDelegate is sufficient.
This makes it possible to use other Activity types like FragmentActivity
and thus supports testing fragments with the FragmentScenario API.


arnaudgiuliani:
Good catch 👌

arnaudgiuliani (approved):

PR #1288, opened 2022-02-09
Use `Set` instead of `List` to handle Included Modules

marcellogalhardo:


marcellogalhardo:
Pull Request closed in favor of @arnaudgiuliani tailrec implementation.

PR #1296, opened 2022-02-18
Improve `includes` docs

marcellogalhardo:
Review `includes` documentation: rephrase some parts, add use cases explanation and make the examples a little more complex to show what are the benefits of the new feature.

### DIFF
--- a/docs/reference/koin-core/dsl.md
+++ b/docs/reference/koin-core/dsl.md
@@ -120,9 +120,15 @@ How does it work? Let's take some modules, and we include modules in `parentModu
 
 ```kotlin
 // `:feature` module
-internal val childModule1 = module { /* Other definitions here. */ }
-internal val childModule2 = module { /* Other definitions here. */ }
-public val parentModule = module { includes(childModule1,childModule2) }
+val childModule1 = module {
+    /* Other definitions here. */
+}
+val childModule2 = module {
+    /* Other definitions here. */
+}
+val parentModule = module {
+    includes(childModule1, childModule2)
+}
 
 // `:app` module
 startKoin { modules(parentModule) }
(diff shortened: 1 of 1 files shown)


docs/reference/koin-core/dsl.md
```diff
@@ -110,61 +110,43 @@ Within this option lambda, you can specify the following options:
...
-val childModule1 = module {
-    // other definitions here
-}
-val childModule2 = module {
-    // other definitions here
-}
+// `:feature` module
+internal val childModule1 = module { /* Other definitions here. */ }
+internal val childModule2 = module { /* Other definitions here. */ }
+public val parentModule = module { includes(childModule1,childModule2) }
```
arnaudgiuliani, line 125: Why do you need to add visibility modifier here? (internal, public)

docs/reference/koin-core/dsl.md
```diff
@@ -110,61 +110,43 @@ Within this option lambda, you can specify the following options:
...
-}
-val featureModule2 = module {
-    includes(dataModule)
-    // other definitions here
-}
+// :feature module
+internal val dataModule = module { /* Other definitions here. */ }
+internal val domainModule = module { /* Other definitions here. */ }
+public val featureModule1 = module { includes(domainModule, dataModule) }
+public val featureModule2 = module { includes(domainModule, dataModule) }
```
arnaudgiuliani, line 146: same I wouldn't use any visibility modifier

arnaudgiuliani:
Just a comment on visibility modifiers in samples

docs/reference/koin-core/dsl.md
marcellogalhardo: That was intentional. It showcases how you can play with visibility modifiers to keep many Koin modules internal while exposing a single "feature module" - that has major implications on how you expose a Gradle module public API.

It is a little hard to explain, as `Module` is overused. Let me give you an example:

image

- Green: Gradle modules.
- Blue: Kotlin files.

As you can note above, `:auth` Gradle module contains multiple Koins modules, but only the `AuthModule` is exposed public while all others are now implementation details. That should give fine control over what Koin module is exposed while keeping the not public ones as `included` modules.

What do you think?

docs/reference/koin-core/dsl.md
marcellogalhardo: Please, see [here](https://github.com/InsertKoinIO/koin/pull/1296#discussion_r823506557).

docs/reference/koin-core/dsl.md
arnaudgiuliani: I see. This is a very good example, but I still wonder if it can go in the "generic" doc 🤔 
It would be a great article for sure.

marcellogalhardo: Right, I will remove it for now.

arnaudgiuliani: 👍

arnaudgiuliani:
@marcellogalhardo I can invite you to publish an article on koin-developers hub if you want.

docs/reference/koin-core/dsl.md
marcellogalhardo: @arnaudgiuliani changes done.

arnaudgiuliani (approved):

PR #1304, opened 2022-03-02
Fix a bug where source jar for koin-ktor and other kotlin projects is empty 

gandhiprachi:
Check the sources jar below.  It's empty as the source code for `koin-ktor` is in `src/main/kotlin` and not `src/main/java`
https://repo1.maven.org/maven2/io/insert-koin/koin-ktor/3.2.0-beta-1/koin-ktor-3.2.0-beta-1-sources.jar
this means we can't attach sources for `koin-ktor` in the ide

### DIFF
--- a/gradle/publish-to-central.gradle
+++ b/gradle/publish-to-central.gradle
@@ -19,7 +19,7 @@ if (!project.tasks.findByName('sourcesJar')) {
         if (pluginManager.hasPlugin('com.android.library')) {
             from android.sourceSets.main.java.srcDirs
         } else {
-            from sourceSets.main.java.srcDirs
+            from sourceSets.main.allSource.srcDirs
         }
     }
 }


arnaudgiuliani:
interesting 🤔 On maven central, the Jar file is filled with content. Not sure it's empty then.

gandhiprachi:
> interesting 🤔 On maven central, the Jar file is filled with content. Not sure it's empty then.
I'm not referring to the jar file that contains the `binaries/classes`. I'm referring to the jar file that contains `sources`. 
[This](https://repo1.maven.org/maven2/io/insert-koin/koin-ktor/3.2.0-beta-1/koin-ktor-3.2.0-beta-1-sources.jar) file should have the source code. I just see the `META-INF` folder. The size is also much smaller than expected

arnaudgiuliani:
ok about the sources :)

arnaudgiuliani:
7e28de47be7545de250f351a487d59f7d8551511 ported in https://github.com/InsertKoinIO/koin-ktor

PR #1306, opened 2022-03-02
Add support for scope linking in checkModules

Pfoerd:
Enables "simulation" of linked scopes in `checkModules`. Thus, we don't have to explicitly provide default values for all of the injected instances coming from other scopes.

```kotlin
val myModule = module {
    scope(named("scope1")) {
        scoped { ComponentA() }
    }
    scope(named("scope2")) {
        scoped { ComponentB(get()) }
    }
}

scope2.linkTo(scope1)
scope2.get<ComponentB>()
```

```kotlin
@Test
fun `test DI modules`(){
    koinApplication {
        modules(myModule)
        checkModules(){
            withScopeLink(named("scope2"), named("scope1"))
        }
    }
}
```


arnaudgiuliani:
Interesting, thanks.

arnaudgiuliani:
Good job

PR #1309, opened 2022-03-09
Update import in Java documentation

GaspardPO:
The Java example uses a decrecated import and is not working.

Real import is this file[org.koin.java.KoinJavaComponent.kt](https://github.com/InsertKoinIO/koin/blob/main/core/koin-core/src/jvmMain/kotlin/org/koin/java/KoinJavaComponent.kt), not a "standalone" file.


arnaudgiuliani:
thanks @GaspardPO, Seems to be a conflict 🙏

arnaudgiuliani:
don't hesitate to reopen 👍

PR #1316, opened 2022-03-21
Add non-inlined functions to Module.

hqzxzwb:
This is an improvement in two ways:
- It provides better inter-operability for native code. Previous inline functions with reified type parameter cannot be called from Swift. Newly added functions allows users to start koin application from native app module in native code.
- It avoids inlining too much bytecode into caller side. After this change and a caller-side recompilation, only a simple call to non-inlined function is inlined to caller side. Previously the whole function logic detail is copied anywhere the function gets called.


arnaudgiuliani:
Thanks for your contribution. Did you check the `constructor DSL`, to help build directly a target class?

arnaudgiuliani:
Yes I see. Let`s add things in the Module API.

arnaudgiuliani:
your proposal has been ported back as Extension on native side. Let's see how it goes.

Thanks for your contribution 🙏

Issue #1242, opened 2021-12-06
java.lang.NoSuchMethodError with kotlin 1.6


krottv:
**Describe the bug**
```
java.lang.NoSuchMethodError:
No static method toDouble-impl(JLjava/util/concurrent/TimeUnit;)D in class Lkotlin/time/Duration; or its super classes (declaration of 'kotlin.time.Duration' appears in
```

**To Reproduce**
1. Set kotlin 1.6 in project
2. Call 

```
    startKoin {

            if (BuildConfig.DEBUG)
                androidLogger()
                
            // declare modules
            androidContext(this@App)
            modules(...)
     }
```

**Koin project used and used version (please complete the following information):**
 [e.g]: `koin-core version 3.1.4`

egorikftp:
Duplicate of https://github.com/InsertKoinIO/koin/issues/1188

Jesil-OT:
works for me, Thanks

Jesil-OT:
works for me, Thanks

tomtaila:
Yeah I have the same issue

Arman10000:
The method helped me
`startKoin { androidLogger(Level.ERROR) }`
Or not to use `androidLogger()`

arnaudgiuliani:
is outdated with Koin 3.2.0.

vuuygur:
add where ?

arnaudgiuliani:
Time API has been ported back. Since Koin 3.2 we are less impacted by Kotlin upgrades

Issue #1216, opened 2021-11-16
androidLogger() doesn't work with Kotlin 1.6


Tuccro:
>     java.lang.NoSuchMethodError: No static method toDouble-impl(JLjava/util/concurrent/TimeUnit;)D in class Lkotlin/time/Duration; or its super classes (declaration of 'kotlin.time.Duration' appears in /data/data/com.example.app/code_cache/.overlay/base.apk/classes18.dex)
>         at org.koin.core.time.MeasureKt.measureDuration(Measure.kt:32)
>         at org.koin.core.KoinApplication.modules(KoinApplication.kt:59)

Application crashes on the start after a set of logger in Application class with the default parameter "level: Level = Level.INFO":

> startKoin {
>             androidLogger()

I think it's obvious, the problem is at **org.koin.core.time.MeasureKt.measureDuration(Measure.kt:32)** and becomes just in case when the logging level is lower than Level.ERROR.

egorikftp:
Duplicate of https://github.com/InsertKoinIO/koin/issues/1188

Issue #1188, opened 2021-10-21
Does not support Kotlin 1.6.0


Nek-12:
**Describe the bug**
Latest koin version produces` NoSuchMethodError`s at runtime when using kotlin 1.6.0-RC

**To Reproduce**
1. Add latest stable koin dependency
2. Use kotlin 1.6.0-RC.
3. App crashes on startup in 100% cases

**Expected behavior**
The app is able to run without crashing. 

**Koin project used and used version (please complete the following information):**
 [e.g]: `koin version 3.1.2`, kotlin 1.6.0-RC
 
 The surest way to get a runtime error is to call a deleted method toDouble() of kotlin.duration that was removed in 1.6.0 by using 
```
startKoin {
            androidContext(applicationContext)
            androidLogger() //This one is causing the crash
            modules(/* blah */)
        }
```

Stacktrace:
```
java.lang.NoSuchMethodError: No static method toDouble-impl(JLjava/util/concurrent/TimeUnit;)D in class Lkotlin/time/Duration; or its super classes (declaration of 'kotlin.time.Duration' appears in /data/app/~~RAEJCWfLUZeoA5jJfJMmtA==/com.nek.test.debug-jvqFgZyETecELyVm71SR5A==/base.apk!classes16.dex)
        at org.koin.core.time.MeasureKt.measureDuration(Measure.kt:32)
        at org.koin.core.KoinApplication.modules(KoinApplication.kt:59)
        at org.koin.core.KoinApplication.modules(KoinApplication.kt:50)
        at com.nek.test.TestApplication$onCreate$1.invoke(TestApplication.kt:30)
        at com.nek.test.TestApplication$onCreate$1.invoke(TestApplication.kt:27)
        at org.koin.core.context.GlobalContext.startKoin(GlobalContext.kt:64)
        at org.koin.core.context.DefaultContextExtKt.startKoin(DefaultContextExt.kt:31)
        at com.nek.test.TestApplication.onCreate(TestApplication.kt:27)
    //bunch of junk    
```

Lesson: do not use experimental APIs

franciscofranco:
Hi. Any update on this issue? Kotlin 1.6.0 has been released to production and this crash happens, making it impossible to upgrade Kotlin...
Thanks.

SerggioC:
same thing. 
[...]

Issue #1076, opened 2021-04-27
 java.lang.NoSuchMethodError: No virtual method elapsedNow-UwyO8pc()D in class Lkotlin/time/TimeMark


ligi:
**Describe the bug**

After updating to Kotlin 1.5.0 I am getting this runtime error:

```
04-27 15:37:54.756  6429  6429 E AndroidRuntime: java.lang.NoSuchMethodError: No virtual method elapsedNow-UwyO8pc()D in class Lkotlin/time/TimeMark; or its super classes (declaration of 'kotlin.time.TimeMark' appears in /data/app/org.walleth-rBcMmS3wfGt-FnuS1dhI1g==/base.apk)
04-27 15:37:54.756  6429  6429 E AndroidRuntime: 	at org.koin.core.time.MeasureKt.measureDuration(Measure.kt:63)
04-27 15:37:54.756  6429  6429 E AndroidRuntime: 	at org.koin.core.KoinApplication.modules(KoinApplication.kt:63)
04-27 15:37:54.756  6429  6429 E AndroidRuntime: 	at org.walleth.App$onCreate$1.invoke(App.kt:142)
04-27 15:37:54.756  6429  6429 E AndroidRuntime: 	at org.walleth.App$onCreate$1.invoke(App.kt:139)
04-27 15:37:54.756  6429  6429 E AndroidRuntime: 	at org.koin.core.context.GlobalContext.startKoin(GlobalContext.kt:65)
04-27 15:37:54.756  6429  6429 E AndroidRuntime: 	at org.koin.core.context.DefaultContextExtKt.startKoin(DefaultContextExt.kt:31)
04-27 15:37:54.756  6429  6429 E AndroidRuntime: 	at org.walleth.App.onCreate(App.kt:139)
04-27 15:37:54.756  6429  6429 E AndroidRuntime: 	at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1182)
04-27 15:37:54.756  6429  6429 E AndroidRuntime: 	at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6460)
04-27 15:37:54.756  6429  6429 E AndroidRuntime: 	at android.app.ActivityThread.access$1300(ActivityThread.java:219)
04-27 15:37:54.756  6429  6429 E AndroidRuntime: 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1859)
[...]

PR #1330, opened 2022-04-17
Avoid using ExperimentalTime API

pubiqq:
Koin breaks down after every minor update of Kotlin because of the use of the experimental time api in logging, it's just ridiculous.

Here is a fix that replaces the use of experimental functions with the use of stable ones.

Prevents the recurrence of issues like:
- https://github.com/InsertKoinIO/koin/issues/1242
- https://github.com/InsertKoinIO/koin/issues/1216
- https://github.com/InsertKoinIO/koin/issues/1188
- https://github.com/InsertKoinIO/koin/issues/1076
- https://github.com/InsertKoinIO/koin/issues/917
- https://github.com/InsertKoinIO/koin/issues/871
- https://github.com/InsertKoinIO/koin/issues/847

### DIFF
--- a/core/koin-core/src/commonMain/kotlin/org/koin/core/time/Measure.kt
+++ b/core/koin-core/src/commonMain/kotlin/org/koin/core/time/Measure.kt
@@ -15,28 +15,38 @@
  */
 package org.koin.core.time
 
-import kotlin.time.DurationUnit
-import kotlin.time.ExperimentalTime
 import kotlin.time.measureTime
 import kotlin.time.measureTimedValue
 
-
 /**
- * Measure functions
+ * Executes the [code] and returns the duration of execution in milliseconds.
  *
- * @author Arnaud Giuliani
+ * TODO: Use [measureTime] when it becomes stable.
  */
+internal fun measureDuration(code: () -> Unit): Double {
+    val timeSource = getTimeSource()
 
-@OptIn(ExperimentalTime::class)
-fun measureDuration(code: () -> Unit): Double {
-    return measureTime(code).toDouble(DurationUnit.MILLISECONDS)
+    val start = timeSource.mark()
+    code()
+    val end = timeSource.mark()
+
+    return (end - start) / 1_000_000.0
 }
 
 /**
- * Measure code execution and get result
+ * Executes the [code] and returns a pair of values - the result of the function execution
+ * and the duration of execution in milliseconds.
+ *
+ * TODO: Use [measureTimedValue] when it becomes stable.
  */
-@OptIn(ExperimentalTime::class)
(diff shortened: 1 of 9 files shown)


arnaudgiuliani:
I agree with you, this time API is breaking updates since too much time. For 3.2 we plan to rework this also, to avoid such breaking. 

Let me check how our solution is going with yours.

pubiqq:
Sure. Feel free to ask me if you have any questions about the changes made.

pubiqq:
> Your feedback is welcome

Well...
- [`self`](https://github.com/InsertKoinIO/koin/blob/0a10adc366ff18c378e21e6ee31a012c50d8baa8/core/koin-core/src/jsMain/kotlin/org/koin/mp/PlatformTimeToolsJS.kt#L7) is not defined in Node.js, so your change doesn't work properly.
- `measureDuration` and `measureTime` are the same thing, just one function is marked with opt-in `@KoinInternalApi`, and the other is not, something obviously superfluous.
- Pair `measureDurationForResult`/`measureTimedValue` has the same problem + `measureDurationForResult` does redundant actions (destructures the result value and creates the same one).
- `performance.now()` is available in WebView for Android SDK 19 and above, but for `koin-android` you declare support for Android SDK 14 and above. It may seem it's about two different things, but since you declare support for Android SDK 14, you admit that such devices exist, so you need to ensure that the library works on them as well.
- File naming leaves much to be desired, like `KoinPlatformTimeTools` class (ok, assume) inside `PlatformTimeToolsJVM.kt` file (you are already inside the `jvm` target, you don't need to mention it again with a suffix)  inside `mp` folder (in short, it's weird to have a "**multi**platform" folder inside a folder associated with a **specific** platform) inside `jvm` target.
- The code is not formatted everywhere, use the "Reformat Code" command or the appropriate keyboard shortcut (for Android Studio: `Ctrl + Alt + L` on Windows and Linux, `⌥⌘L` on Mac). If you constantly forget about it, or you don't want to do it manually, you can configure the environment to do this before commit (for Android Studio: "File" -> "Settings" -> "Version Control" -> "Commit" -> group "Before Commit" -> select the necessary actions: reformat, cleanup and so on). Also, you can delegate this work to your CI platform (there is no specific advice here, it all depends on the CI platform and your development process).

To be honest, I would still recommend using my PR. 🙃

arnaudgiuliani:
My bad. I took all the code on JS side:

https://github.com/InsertKoinIO/koin/blob/3.2.x/core/koin-core/src/jsMain/kotlin/org/koin/mp/PlatformTimeTools.kt
https://github.com/InsertKoinIO/koin/blob/3.2.x/core/koin-core/src/jsMain/kotlin/org/koin/core/time/TimeSource.kt

PR #1339, opened 2022-05-11
Fixed eager instances not working on native targets

remvst:
Noticed this while testing with Kotlin multiplatform. This should fix tests on CI (been failing for a while it seems)


arnaudgiuliani (approved):

remvst:
@arnaudgiuliani do I need to do anything from there? Or is it up to you to include the fix in 3.2.1?

arnaudgiuliani:
@remvst Your commits are not signed. Can you check?

PR #1359, opened 2022-06-06
Add Constructor DSL up to 22 parameters

### LINKED ISSUE #1342: Add singleOf for `11..22` params
I use `singleOf(::MyClass)`, but `MyClass` constructor has 11 params.
Version 3.2 has provided `singleOf` with below 11 params

marcellogalhardo:
Add constructor DSL up to 22 parameters for: `new`, `singleOf`, `factoryOf`, `scopedOf`, `viewModelOf`, `workerOf`, `fragmentOf`. It also simplifies the constructor's internal implementation by inlining `withOptions` (if empty, the inlined `options` parameter should not create an instance and thus performance should not be impacted).

Note 22 was chosed due to the Kotlin STD Lib using that number. To keep it synchronized with the STD Lib, we won't be adding more than 22.

Fixes #1342.

### DIFF
--- a/android/koin-android/src/main/java/org/koin/androidx/fragment/dsl/FragmentOf.kt
+++ b/android/koin-android/src/main/java/org/koin/androidx/fragment/dsl/FragmentOf.kt
@@ -18,6 +18,9 @@
 package org.koin.androidx.fragment.dsl
 
 import androidx.fragment.app.Fragment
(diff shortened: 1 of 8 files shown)


arnaudgiuliani:
thanks @marcellogalhardo for your contribution 👍

Issue #213, opened 2018-08-30
Manually providing the ViewModelProvider to `by viewModel()`


miquelbeltran:
**Is your feature request related to a problem? Please describe.**

I maintain a little library that adds ViewModel support from AAC to Conductor, and recently got this question: https://github.com/miquelbeltran/conductor-viewmodel/issues/1

```
Have you used Koin at all to inject your ViewModels?
```

I tried and found an issue: Koin relies on the parent Activity or the Fragment to obtain the ViewModelProvider. However, my `Controller` class (a replacement for Fragments by Conductor) also is a ViewModelProvider.

This is the method/line that causes the problem: https://github.com/InsertKoinIO/koin/blob/f71200780dd2018e5fdd6292b0a02772396117bd/koin-android-architecture/src/main/java/org/koin/android/architecture/ext/LifecycleOwnerExt.kt#L166

**Describe the solution you'd like**

A solution would be having an optional paramater in `viewModel()` and `getViewModel()`, being able to pass the `ViewModelProvider` by parameter.

So I can call: `val myViewModel by viewModel(viewModelProvider = getMyViewModelProvider())`

And the `getViewModelByClass` will use that provided `ViewModelProvider` instead.

**Describe alternatives you've considered**

The alternative is to not to use the ViewModel Koin extension and inject the ViewModelFactories instead.

**Target Koin project**

this is an improvement for `koin-android-architecture`

----

Thanks for considering! I would be very happy to implement it and provide a PR with the solution if you think it is worth it. I am happy user of Koin and as well Conductor, and would like to see them working together. :-)

arnaudgiuliani:
Yeah, good to ping it.

arnaudgiuliani:
included in 2.0

Sar777:
wow! thank you!

Sar777:
@arnaudgiuliani one question, could you provide an information how to use the feature in 2.0.0?

arnaudgiuliani:
with frag

arnaudgiuliani:
sorry, false alert.

ViewModelProvider is generated by Koin: https://github.com/InsertKoinIO/koin/blob/master/koin-projects/koin-android-viewmodel/src/main/java/org/koin/android/viewmodel/ViewModelResolution.kt#L33

The need above is to replace this part then?

miquelbeltran:
Ah, as I understand now, we can provide the ViewModelParameters.from as any custom ViewModelProvider we would like to use.

So we can do now:

[...]

PR #1373, opened 2022-06-29
Add viewmodel get key to koin, and add KoinViewModelLazy class

allen-hsu:
Hi, because the ViewModel provider has the function (`ViewModelProvider.get(key, clazz)`) that can let you pass the key to ViewModel provide in order to get a different ViewModel with a different key, I want to raise this PR, and according to this : 
https://github.com/InsertKoinIO/koin/issues/213

Originally I was going to change `ViewModelProvider.resolveInstance`, but I found that the new version of koin uses ViewModelLazy instead, but it is an Android lib, so I tried to build a KoinViewModelLazy, please consider this scenario carefully, I believe that I am not the only one who needs to pass in the key

### DIFF
--- a/android/koin-android-compat/src/main/java/org/koin/android/compat/ScopeCompat.kt
+++ b/android/koin-android-compat/src/main/java/org/koin/android/compat/ScopeCompat.kt
@@ -16,8 +16,8 @@
 package org.koin.android.compat
 
 import androidx.lifecycle.ViewModel
-import androidx.lifecycle.ViewModelLazy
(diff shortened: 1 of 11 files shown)


arnaudgiuliani:
The VIewModel API has been revamped with latest lifecycle library in 2.5.1. Key property can be accessed in generic API.
https://github.com/InsertKoinIO/koin/pull/1459

Thanks for your proposal. Don't hesitate to give feedback on `koin-android` 3.3.0

PR #1382, opened 2022-07-13
Add createEagerInstances on context-isolation

rvp-diconium:
It's necessary to call `createEagerInstances()` when working with context-isolation (library / SDK development) or else singletons declared with `single(createdAtStart = true)` will be lazily initiated.


arnaudgiuliani:
thanks for your contribution

Issue #1318, opened 2022-03-22
Provide a way to load modules immediatly for Android instrumented tests


christophehenry:
**Is your feature request related to a problem? Please describe.**
As [the documentation states](https://insert-koin.io/docs/reference/koin-android/instrumented-testing), on Android, Koin is started by the `Application` class which means that loaded modules are the same for every instrumentation tests.

The documentation states that "*`loadModules` and `unloadModules` are often unsafe because the changes are not applied immediately*" which means that modules can't reliebly be swapped — for instance by a mock — for the course of one or a series of specific tests.

It is not clear here if `createEagerInstances()` and `single`s `createdAtStart` parameter provide a partial solution to that problem by always immediatly creating singleton modules.

**Describe the solution you'd like**
`loadModules` and `unloadModules` could provide a `executeImmediatly` for swapping modules with immediate exectution.

**Target Koin project**
Koin

gmk57:
Yes, we're looking for it too!

The topic of proper Koin setup for instrumented tests is discussed for [almost two years](https://github.com/InsertKoinIO/koin/issues/850), but sill lacks a perfect solution.

Part of the problem is probably the vague documentation: "*recommended approach is to add a `module` of your overrides to `modules` used by `startKoin` in the application's `Application` class*".

Where should we do that?

Adding `instrumentedTestModule` in production `Application` is neither possible nor desirable. Docs probably mean using a special `Application` class for tests via [manifest](https://stackoverflow.com/a/53523241) or [test runner](https://github.com/InsertKoinIO/koin/issues/850#issuecomment-773364141), but this is not said clearly.

I find it strange compared to very clear & practical Koin documentation in general (yes, we're looking at you, Dagger ;)).

christophehenry:
> Docs probably mean using a special Application class for tests via [manifest](https://stackoverflow.com/a/53523241) or [test runner](https://github.com/InsertKoinIO/koin/issues/850#issuecomment-773364141), but this is not said clearly.

I think docs mean via test runner, which is what I use. The first problem is this solution seems not described anywhere in Android's documentation (I found my answer on StackOverflow). The second is there is no way to load different modules for different tests. The runner and thus the `Application` is the same for all the instrumented tests. You can't isolate one module for one test, for instance, in order to mock it.

[...]

Issue #850, opened 2020-07-18
how to properly setup android UI tests?


or-dvir:
Hello.
I can't seem to figure out how to properly setup my android UI tests. no metter what configuration i use i get some sort of koin related exception. here is my setup:

version used:
```
implementation "org.koin:koin-androidx-viewmodel:2.1.6"
testImplementation "org.koin:koin-test:2.1.6"
androidTestImplementation "org.koin:koin-test:2.1.6"
```

application class:
```
class MyApplication : Application() {

    companion object {
        val appModule = module {
            viewModel { FragmentUserListsViewModel(get()) }
            viewModel { FragmentNewEditListViewModel(get()) }
            viewModel { FragmentListItemsViewModel() }
            single<RepositoryUserLists> { RepositoryUserListsImpl() }
            factory<RepositoryListItems> { (listId: UUID) -> RepositoryListItemsImpl(listId) }
        }

        val testModule = module(override = true) {
            single<RepositoryUserLists> { RepositoryUserListsImplFake() }
            factory<RepositoryListItems> { (listId: UUID) -> RepositoryListItemsImplFake(listId) }
        }
    }

    override fun onCreate() {
        super.onCreate()

        startKoin {
            androidLogger()
            androidContext(this@MyApplication)
            modules(appModule)
        }
    }
}
```

## test configuration 1:
```
    @get:Rule
    val koinTestRule = KoinTestRule.create {
        //so i can override the real repositories with fake ones
        modules(MyApplication.appModule, MyApplication.testModule)
    }
```
the first test always fails with `java.lang.IllegalStateException: A KoinContext is already started`. all other tests run fine and pass. its not a specific test that fails, but always the first one.
here is the stack trace:

```
java.lang.IllegalStateException: A KoinContext is already started
at org.koin.core.context.KoinContextHandler.register(KoinContextHandler.kt:49)
at org.koin.core.context.ContextFunctionsKt.startKoin(ContextFunctions.kt:36)
at org.koin.core.context.ContextFunctionsKt.startKoin$default(ContextFunctions.kt:35)
at org.koin.test.KoinTestRule.starting(KoinTestRule.kt:25)
at org.junit.rules.TestWatcher.startingQuietly(TestWatcher.java:108)
at org.junit.rules.TestWatcher.access$000(TestWatcher.java:46)
at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:53)
[...]

PR #1409, opened 2022-08-12
Enhance Android Instrumentation tests doc

cmunaro:
I know that some people get frustrated when they try to make Instrumentation test with Koin for the lack of examples. (I was one of them 😅)
I propose a documentation update to make it easier for newcomers to create instrumentation tests with Koin. 🔧
What do you think about it?

Related to https://github.com/InsertKoinIO/koin/issues/1318 and https://github.com/InsertKoinIO/koin/issues/850


arnaudgiuliani:
Good proposal that will help many people 👍

sbarbour-93:
Thanks for updating the docs on this - it's a really useful update. However, I've noticed one issue though in that if the module I try to override expects parameters to be injected in order for the instance to be built then my instrumentation tests fall over when using the new `KoinTestRule` approach with a `DefinitionParameterException`.

Is there any suggestions on this? If it needs a new issue to be opened in order to review please advise.

Thanks!

@arnaudgiuliani

CoolMind:
See https://insert-koin.io/docs/reference/koin-android/instrumented-testing/ or commits here for an updated documentation.

maeceloacm1998:
Muito obrigado, sua solução me ajudou para fazer testes com espresso.

Issue #1446, opened 2022-10-04
Crash on koinNavGraphViewModel after 3.2.2 update


LipeDuoli:
**Describe the bug**
I tried to update my project from version 3.2.0 to 3.2.2.
After migrate to the new [android scope](https://insert-koin.io/docs/reference/koin-android/scope#android-scope-classes) from the previous lazy delegate to the new `createFragmentScope()` All my calls to `by koinNavGraphViewModel` is crashing with the error
```
Trying to access Android Scope on 'HomeFragment{c1f0992} (aeee074c-3115-41a1-b87f-84b9bfb93494)' but scope is not created
```
my code is
```
private val viewModel: HomeViewModel by koinNavGraphViewModel(R.id.navGraph)
```
And if I change to 
```
private val viewModel: HomeViewModel by viewModel()
```
the code do not crash anymore

Other places that I use just `by viewModel()`, continue to work fine

**To Reproduce**
1. Create an fragment with the new android scope implementation (`createFragmentScope()`)
2. Inject an viewModel using `by koinNavGraphViewModel`

**Expected behavior**
scope to be inicialized when `koinNavGraphViewModel` is used

**Koin project used and used version (please complete the following information):**
_Before:_
`koin-android:3.2.0`
`koin-androidx-workmanager:3.2.0`
`koin-androidx-navigation:3.2.0`
`koin-android-compat:3.2.0`

_After:_
`koin-android:3.2.2`
`koin-androidx-workmanager:3.2.2`
`koin-androidx-navigation:3.2.2`
`koin-android-compat:3.2.2`

arnaudgiuliani:
Scope resolution has been fixed 👍

Issue #455, opened 2019-05-17
[2.0.0-GA5] Named instances resolutions are broken 


renaudcerrato:
Just migrated from `rc-2` to `GA5`, but I noticed that named definitions resolution is buggy. Root scope named definitions are hidden by un-named definitions in the current scope:

```
val fooModule = module {
  single(named("NAMED")) { Foo() }

  scope(named("MY_SCOPE")) {
     scoped { Foo() }
  }
} 

val barModule = module {
  // will get un-named Foo from scope instead of named Foo from root ! 
  single { Bar(get<Foo>(named("NAMED"))) }
}

getKoin().getOrCreateScope("scope_id", named( "MY_SCOPE")).get<Bar>()
``` 

The bug is located at [`Scope.findDefinition()`](https://github.com/InsertKoinIO/koin/blob/2.0.0-GA/koin-projects/koin-core/src/main/kotlin/org/koin/core/scope/Scope.kt#L145) since [`BeanRegistry.findDefinition()`](https://github.com/InsertKoinIO/koin/blob/2.0.0-GA/koin-projects/koin-core/src/main/kotlin/org/koin/core/registry/BeanRegistry.kt#L129) will return a definition matching the class (on the current scope !) if there's no named definition into that scope:

```kotlin
fun findDefinition(
            qualifier: Qualifier? = null,
            clazz: KClass<*>
    ): BeanDefinition<*>? =
            qualifier?.let { findDefinitionByName(it.toString()) }
                    ?: findDefinitionByClass(clazz)

```

arnaudgiuliani:
Your scope is shadowing definitions from the global space, which is normal: when you define a scope and resolve from it, we are looking firstly from the scope definitions and then from the global scope.

Apart from that your case here is quite tricky ^^... as I wouldn't advise spreading scoped instances around to leak it?

renaudcerrato:
Mmm... I'm asking for a named instance, and I'm ending up with an unnamed one! It was working perfectly on rc2. Could you read the issue carefully again?

If there's no definitions matching the qualifier on the current scope, we should look into the root scope before looking for a "class-only" definition.

renaudcerrato:
The look-up for definitions should be : `findByName(qualifier) ?: root.findByName(qualifier) ?: findByClass(clazz) ?: root.findByClass(clazz)`

arnaudgiuliani:
yes I see

arnaudgiuliani:
try with `2.0.0-GA6`

[...]

PR #1454, opened 2022-10-18
#1446 koinNavGraphViewModel fix 

M-Olos-MOL:
**Caused by: java.lang.IllegalStateException: Trying to access Android Scope on 'xxxxFragmen' but scope is not created** after Koin update


arnaudgiuliani:
Thanks, already fixed in #455

PR #1465, opened 2022-11-02
Fix race condition in scope resolveInstance

### LINKED ISSUE #1149: java.util.NoSuchElementException when calling getOrNull() while creating an instance of an injected class with parameters
**Describe the bug**
Koin crashes with the following exception when resolving an instance with parameters that has a dependant call to `getOrNull()` (which gets not provided) to create the instance.
```
java.util.NoSuchElementException: ArrayDeque is empty.
    at kotlin.collections.ArrayDeque.removeFirst(ArrayDeque.kt:145)
    at org.koin.core.scope.Scope.resolveInstance(Scope.kt:213)
```

**To Reproduce**
1. Create a definition that
**(a)** gets parameters from `parametersOf()`
**(b)** requires a call to `getOrNull()` to create the instance for this definition
```kotlin
factory { params ->
    MyClass(
        name = params.get(),
        optionalInjectedParameter = getOrNull()
    )
}
```
2. Do **not** provide a definition for the parameter that should be resolved via `getOrNull()`
3. Inject it somewhere
```kotlin
val myInjectedClass = get<MyClass> {
    parametersOf("Some parameter")
}
```

**Whats the problem**
* Koin calls `Scope.resolveInstance()` for `MyClass` and puts the parameters on the `_parameterStack`. It then calls `Scope.resolveValue()` to create an instance of `MyClass`.
* While creating the instance Koin does a call to `getOrNull()` as we defined it to provide the optional parameter if it is available. It calls `Scope.resolveValue()` again for the optional parameter and ends up calling `_parameterStack.clear()` and `throwDefinitionNotFound()` from there. The exception is caught within getOrNull() so it gracefully returns `null` for the optional parameter from there and the instance of `MyClass` gets created correctly in `Scope.resolveInstance()`.
* But afterwards it tries to remove the parameters of `MyClass` with `_parameterStack.removeFirst()`. And since the stack was already cleared from within the `getOrNull()` call it crashes here.

**Expected behavior**
Koin should create the instance correctly and should not fail due to intentional optional parameters.
Please note that it can also be the case that we need to call `getOrNull()` from the init block of the injected class for some reasons and not only provide it as a constructor parameter.

**Koin project used and used version (please complete the following information):**
`koin-core version 3.1.2`

npresseault:
Fix an issue where concurrent calls to koin.get would create a race condition resulting in this stack trace:

```
Fatal Exception: java.util.NoSuchElementException
ArrayDeque is empty.
kotlin.collections.ArrayDeque.removeFirst (ArrayDeque.kt:145) 
kotlin.collections.ArrayDeque.removeFirstOrNull (ArrayDeque.java:157) 
org.koin.core.scope.Scope.resolveInstance (Scope.kt:244) 
org.koin.core.scope.Scope.get (Scope.kt:204)
org.koin.core.scope.Scope.getOrNull (Scope.kt:172)
```

ArrayDeque is not a thread-safe collection, by wrapping calls to parameterStack in a synchronized block it makes sure the race condition cannot happen.

Also added a test that reproduced quite easily the problem.

Closes #1149

### DIFF
--- a/core/koin-core/build.gradle
+++ b/core/koin-core/build.gradle
@@ -49,6 +49,7 @@ kotlin {
             dependencies {
                 implementation kotlin("test-common")
                 implementation kotlin("test-annotations-common")
+                implementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.6.4"
             }
         }
 
--- a/core/koin-core/src/commonMain/kotlin/org/koin/core/scope/Scope.kt
+++ b/core/koin-core/src/commonMain/kotlin/org/koin/core/scope/Scope.kt
@@ -218,13 +218,17 @@ data class Scope(
         val parameters = parameterDef?.invoke()
         if (parameters != null) {
             _koin.logger.log(Level.DEBUG) { "| put parameters on stack $parameters " }
-            _parameterStack.addFirst(parameters)
+            KoinPlatformTools.synchronized(this@Scope) {
+                _parameterStack.addFirst(parameters)
+            }
         }
         val instanceContext = InstanceContext(_koin, this, parameters)
         val value = resolveValue<T>(qualifier, clazz, instanceContext, parameterDef)
         if (parameters != null) {
             _koin.logger.log(Level.DEBUG) { "| remove parameters from stack" }
-            _parameterStack.removeFirstOrNull()
+            KoinPlatformTools.synchronized(this@Scope) {
+                _parameterStack.removeFirstOrNull()
+            }
         }
         return value
     }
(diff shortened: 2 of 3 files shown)


core/koin-core/build.gradle
```diff
@@ -49,6 +49,7 @@ kotlin {
             dependencies {
                 implementation kotlin("test-common")
                 implementation kotlin("test-annotations-common")
+                implementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.6.4"
```
egorikftp, line 52: Why not `testImplementation`

takke:
Great job!

This PR solves exactly the problem I have been chasing for the past few days.
Hundreds of people are encountering crashes on my app every day.

I am waiting for it to be merged and released ASAP.

takke (approved):

arnaudgiuliani:
Let me check it. Need to see perf impact also.

arnaudgiuliani:
Good job @npresseault 👍

PR #1478, opened 2022-11-18
Updating readme with the correct class names

hardik-trivedi:
There is no WorkFactory class. It should be WorkerFactory. I also renamed example class names accordingly.


arnaudgiuliani:
thanks

PR #1484, opened 2022-12-07
Update annotations.md - Fixed a Typo

broken-bytes:
There was an additional `"` which caused an error.


arnaudgiuliani:
this file shouldn't be there. I will remove it. It's in the other repo.

PR #1503, opened 2023-01-12
Fix conditions for resolving a value by the scope's source

Pfoerd:
Currently a value is resolved by the scope's source even if the class of the definition is not equal to the class of the source value and even if a qualifier is actually defined. 

This leads to undesired behavior in some cases, e.g. when the source of a scope is an Android Activity and the Application Context should be injected from that scope:

```
// in Android Application onCreate
startKoin {
  androidContext(context) // context is Application
  modules(
    module {
      scope<MyActivity> {
        scoped { MyPresenter(androidContext()) }
      }
    }
  )
}
```

```
class MyActivity : ScopeActivity() {
    val presenter : MyPresenter by inject() // WRONG: MyPresenter gets this activity injected as Context instead of the ApplicationContext as it is the scope's source and is an instance of Context
}
```

This PR fixes that behavior.

### DIFF
--- a/core/koin-core/src/commonMain/kotlin/org/koin/core/scope/Scope.kt
+++ b/core/koin-core/src/commonMain/kotlin/org/koin/core/scope/Scope.kt
@@ -251,7 +251,7 @@ data class Scope(
         ?: run {
             _koin.logger.debug("|- ? t:'${clazz.getFullName()}' - q:'$qualifier' look at scope source" )
             _source?.let {
-                if (clazz.isInstance(it)) {
+                if (it::class == clazz && qualifier == null) {
                     _source as? T
                 } else null
             }


Pfoerd:
@arnaudgiuliani any comments on this? :)

arnaudgiuliani:
Need to check side effects yes 👍

arnaudgiuliani:
Seems ok. Let's go and keep on eye on it for 3.4

giangpham96:
This PR seems to introduce a new issue, I filed it [here](https://github.com/InsertKoinIO/koin/issues/1595)

giangpham96:
I think a better fix for this issue would be to add an internal qualifier when binding the context using `androidContext()`

arnaudgiuliani:
@Pfoerd investigating more, in this case, @Pfoerd you should use `androidApplication()` instead of `androidContext()` to resolve the Application context, and not the closest `Context` 

I revert back the patch to use type match with `isInstance` 
```kotlin
 if (clazz.isInstance(source) && qualifier == null) {
                        _source as? T
                    }
```

I will check also the documentation to add this case, to et people know the difference between `androidContext()`, `androidApplication()` and `get()` in the DSL.

PR #1575, opened 2023-05-03
Fix minor things in Annotations docs

zsmb13:


marcellogalhardo (approved):

zsmb13:
@marcellogalhardo That patch is only in the `core/3.4.1` branch, should I change this PR to target that branch?

pedrofsn:
I think so because branch master is already protected

pedrofsn (approved):

Issue #1557, opened 2023-04-10
Robolectric App tests get a closed Koin in composables


yschimke:
**Describe the bug**

Using Koin with Robolectric, fails with 

```
Scope '_root_' is closed
org.koin.core.error.ClosedScopeException: Scope '_root_' is closed
	at org.koin.core.scope.Scope.resolveInstance(Scope.kt:221)
	at org.koin.core.scope.Scope.get(Scope.kt:210)
	at org.koin.androidx.viewmodel.factory.KoinViewModelFactory.create(KoinViewModelFactory.kt:25)
	at androidx.lifecycle.ViewModelProvider.get(ViewModelProvider.kt:187)
	at androidx.lifecycle.ViewModelProvider.get(ViewModelProvider.kt:153)
	at org.koin.androidx.viewmodel.GetViewModelKt.resolveViewModel(GetViewModel.kt:44)
```

When it fails, these return different Koin instances in the composable

```
KoinPlatformTools.defaultContext().get() -> the fresh instance
LocalKoinApplication.current -> the instance from the first test
```

**To Reproduce**
1. Checkout https://github.com/joreilly/Confetti/, specifically this PR https://github.com/joreilly/Confetti/pull/580
2. Run tests in wearApp/src/test
3. Disable the overrides in MainActivity
4. Run tests again

** workaround **

In main app composable, override

https://github.com/joreilly/Confetti/pull/580/files#diff-0548d0af1b708341bc3c142fbf71b1270fb90f0653f791f2eea9e92867d2e99f

```
        setContent {
            ConfettiApp(navController, intent)
            // This shouldn't be needed, but allows robolectric tests to run successfully
            // TODO remove once a solution is found or a fix in koin?
            CompositionLocalProvider(
                LocalKoinScope provides KoinPlatformTools.defaultContext().get().scopeRegistry.rootScope,
                LocalKoinApplication provides KoinPlatformTools.defaultContext().get()
            ) {
...
```

**Expected behavior**
Tests should pass as there is a fresh Koin instance for the current test

**Koin project used and used version (please complete the following information):**
3.4.0

chankeiro:
I am experiencing the same issue with koin-androidx-compose:3.4.3 and . With 3.4.1 + koin-android 3.4.0 everything works fine. This is the error I get:
`Caused by org.koin.core.error.ClosedScopeException: Scope '_root_' is closed
       at org.koin.core.scope.Scope.resolveInstance(Scope.kt:221)
[...]

PR #1586, opened 2023-05-17
Avoid retaining outdated LocalKoinApplication/LocalKoinScope

jjkester:
The `CompositionLocal`s that Koin uses only work when the (`GlobalContext`) Koin application is not changed at runtime, see #1557 

#### Option 1: The `KoinApplication` composable

The Koin application is looked up in the `Context` tree and the Koin `CompositionLocal`s are provided with/from this value. This (still) requires this function to be called around the first time Koin is used from Compose. A practical place for this is the `setContent` function. This means that (boilerplate) code needs to be added to every Activity or Fragment using Compose, and every test case where a Composable is tested in isolation.

Benefits:
- Works in Kotlin Compose Multiplatform
- Follows principles of already existing APIs

Drawbacks:
- Use is required to keep Compose and Context based APIs in sync, easy to forget to add

#### Option 2: Throwing `UnknownKoinContext` in the default value factory

By throwing an exception as default factory for the `CompositionLocal`s we signal that there is no explicit value set. This will trigger the default lookup behavior that used to be the result of the factory function, as long as the appropriate functions `getKoin()` and `getKoinScope()` are used. By using the internal Compose API we are able to catch the exception to run the lookup code. We remember the result of the try/catch block to ensure we only incur the overhead of the exception once per `getKoin()` call per composition.

The default value of a `CompositionLocal` is singleton for the JVM process; when remembering a value as done in this approach the specific value is tied to the composition where it is used.

Benefits:
- Developers do not have to add the `KoinApplication` composable as compared to option 1

Drawbacks:
- Exceptions means a performance hit
- Uses an internal Compose API that we are not supposed to call

#### Other options

- Always retrieve the Koin application from the Android Context on Android
  - This means a specific Android implementation while the MPP code stays the same
- Reuse the initially referenced objects in Koin
  - This means keeping the same global `Koin` and `Scope` objects, even when a different Koin application is started using the test rule
  - Unsure about viability
- Ask Compose devs nicely to create a `CompositionLocal` variant that evaluates the default value expression every time instead of using `lazy` (and wait a long time for them to implement it)

### Open tasks

- [ ] Investigate alternatives
- [ ] Investigate integrations with test frameworks for testing standalone composables
- [ ] Analyze performance of exception strategy
- [ ] Add test cases
  - [ ] Looking up the right Koin application from the Android context
  - [ ] Resolving #1557 for Robolectric and Android instrumented tests

### Open questions to maintainer(s)

- Where to place Robolectric and instrumented tests?
- What is the preferred solution?

### DIFF
--- a/compose/koin-androidx-compose/src/main/java/org/koin/androidx/compose/KoinApplication.kt
+++ b/compose/koin-androidx-compose/src/main/java/org/koin/androidx/compose/KoinApplication.kt
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2017-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@file:OptIn(KoinInternalApi::class)
+
+package org.koin.androidx.compose
+
+import android.app.Application
+import android.content.ComponentCallbacks
+import android.content.Context
+import android.content.ContextWrapper
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.remember
+import androidx.compose.ui.platform.LocalContext
+import org.koin.android.ext.android.getKoin
+import org.koin.compose.LocalKoinApplication
+import org.koin.compose.LocalKoinScope
+import org.koin.core.annotation.KoinInternalApi
+import org.koin.core.component.KoinComponent
+
+/**
+ * Provide active Koin application from Android [Context] to Compose
+ *
+ * @param content - following compose function
+ *
+ * @author Jan-Jelle Kester
+ */
+@Composable
+fun KoinApplication(content: @Composable () -> Unit) {
+    val context = LocalContext.current
+    val koinApplication = remember(context) {
+        context.findContextForKoin().getKoin()
+    }
+    CompositionLocalProvider(
+        LocalKoinApplication provides koinApplication,
+        LocalKoinScope provides koinApplication.scopeRegistry.rootScope,
+        content = content
+    )
+}
+
+/**
+ * Find the [KoinComponent] in the Context tree
+ */
+private fun Context.findContextForKoin(): ComponentCallbacks {
+    var context = this
+    while (context is ContextWrapper) {
+        if (context is KoinComponent && context is ComponentCallbacks) return context
+        context = context.baseContext
+    }
+    return applicationContext as Application
+}
--- a/compose/koin-androidx-compose-navigation/src/main/java/org/koin/androidx/compose/navigation/NavViewModel.kt
+++ b/compose/koin-androidx-compose-navigation/src/main/java/org/koin/androidx/compose/navigation/NavViewModel.kt
@@ -18,11 +18,12 @@
 package org.koin.androidx.compose.navigation
 
 import androidx.compose.runtime.Composable
-import androidx.lifecycle.*
+import androidx.lifecycle.ViewModel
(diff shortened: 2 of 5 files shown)


compose/koin-androidx-compose/src/main/java/org/koin/androidx/compose/KoinApplication.kt
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Provide active Koin application from Android [Context] to Compose
+ *
+ * @param content - following compose function
+ *
+ * @author Jan-Jelle Kester
+ */
+@Composable
+fun KoinApplication(content: @Composable () -> Unit) {
```
arnaudgiuliani, line 42: The `koin-compose` module has already a `KoinApplication` function. Do we need a "special" one here?

jjkester: For this specific use-case, yes.

The `KoinApplication` function that already exists _creates_ a new application and starts it. For this one, the assumption is that externally (likely in the Android `Application` class) an application was already created. This is an Android specific implementation because it uses the Android `Context` to find the relevant Koin application/scope. This is all not implemented for multiplatform Compose (as it works differently), so this code cannot live in `koin-compose`.

In short, this is needed to work around the issue that the composition local defaults are never changing between test runs, ensuring that we have a way of setting them to the correct value.

Personally I don't like this workaround, since we are introducing production code purely for test purposes, but with the way composition locals in Compose are built and the way Koin uses them (for composition-based scopes) I could not think of a good alternative.

arnaudgiuliani:
Hey @jjkester 

I've ported your work on 61a88bbf79d593c9ed777f5b1acb07caa5e6db2e

Lets see how it goes with `KoinContext` and `KoinAndroidContext`

jjkester:
@arnaudgiuliani Thanks for the heads-up, do to a busy time at work and the summer holidays I completely forgot about this.

Your commit looks good at a first glance, I'll definitely replace the workaround in our project with this approach once released.

I'd still would like to write a regression test for this, earlier I could not find a place in the codebase where to put it. Do you have any thoughts on this?

I'd like to make the tests as "real" as possible by using the Koin test utilities on a dummy project, of course ensuring that the tests fail without applying `KoinScope`/`KoinAndroidScope`.

arnaudgiuliani:
@jjkester there is an issue on that: https://github.com/InsertKoinIO/koin/issues/1668

I believe I could make access to scope like this:

```kotlin
fun currentKoinScope(): Scope = currentComposer.run {
    try {
        consume(LocalKoinScope)
    } catch (_: UnknownKoinContext) {
        val ctx = getKoinContext()
        ctx.warnNoContext()
        getKoinContext().scopeRegistry.rootScope
    }
}


fun rememberCurrentKoinScope(): Scope = currentComposer.run {
    remember {
        try {
            consume(LocalKoinScope)
        } catch (_: UnknownKoinContext) {
            val ctx = getKoinContext()
            ctx.warnNoContext()
            getKoinContext().scopeRegistry.rootScope
        }
    }
}
```

With or without remember for  LocalKoinScope, to help reevaluate scope if needed

jjkester:
All the more reason to create some extensive test cases for this kind of thing (it is just fairly complicated!)

Have not had time yet to delve into the mentioned issue, but from the surface it seems to make sense. Because the composition local is accessed from within `remember` it won't invalidate the computed result (since it is not passed as a parameter).

Besides plenty of regression tests I'd also advise to check the performance without `remember` (not sure how though). Maybe there's a more low-level API that can be leveraged to claw back some of the performance.

From a performance standpoint you'd want to avoid the `UnknownKoinContext` exception being thrown very often on your main thread.

arnaudgiuliani:
> All the more reason to create some extensive test cases for this kind of thing (it is just fairly complicated!)
> 
> Have not had time yet to delve into the mentioned issue, but from the surface it seems to make sense. Because the composition local is accessed from within `remember` it won't invalidate the computed result (since it is not passed as a parameter).
> 
> Besides plenty of regression tests I'd also advise to check the performance without `remember` (not sure how though). Maybe there's a more low-level API that can be leveraged to claw back some of the performance.
> 
> From a performance standpoint you'd want to avoid the `UnknownKoinContext` exception being thrown very often on your main thread.
> 
> In my project we did not run into this issue since we do not use `koinInject`.

I went with this PR https://github.com/InsertKoinIO/koin/pull/1723

I've tried manual benchmarking around the internals, with local measure calls. Seems to be faster without remember.

Issue #1578, opened 2023-05-08
New viewmodel instances are not provided when different parameters are specified during injection


snaum47:
**Describe the bug**
Inside our fragment we inject viewmodels with different parameters provided each time. 
The 1st instance with its parameters is provided on all future injections of the viewmodel in the same fragment.

**To Reproduce**
```
// the module
fun appModule() = module {
  viewModel { parameters ->
    SomeViewModel(parameters.get())
  }
}

// the viewmodel
class SomeViewModel(
  val id: UUID
): ViewModel() {

  fun printId() {
    println("Printing SomeViewModel $id")
  }

}
class SomeFragment : Fragment() {
  override fun onResume() {
    super.onResume()
    initialize()
    initialize()
  }

  private fun initialize() {
    val viewModel: SomeViewModel by viewModel {
      parametersOf(UUID.randomUUID())
    }
    viewModel.printId()
  }
}
```
What will be printed is:
```
Printing SomeViewModel 2e7d1d81-3506-49c6-845b-6f926ab8fa86
Printing SomeViewModel 2e7d1d81-3506-49c6-845b-6f926ab8fa86
```

**Expected behavior**
What's expected is that there should be a new viewmodel instance for the 2nd initialize call. The UUID for the 2nd print should be different. 

**Koin project used and used version (please complete the following information):**
 `koin-android version 3.4.0`

**Additional moduleDefinition**
Add any other moduleDefinition about the problem here.

arnaudgiuliani:
yes, it's intended here as we don't have an identifier depending on parameters:
![image](https://github.com/InsertKoinIO/koin/assets/2360688/21954973-79ba-4fc7-b363-8044dcde338c)

the ViewModel is held by WM Lifecycle API.  Before getting into patch consideration, look on your side if VM is the right definition of support. You could use factory to do what you want here.

p-fischer:
From a user point of view, parameters in the `viewModel{}`construct are most likely used as constructor argument of a view model constructor. A view model constructor parameter is scoped to the lifetime of the view model instance. 
So I'd expect that the `viewModel{}` construct would create/resolve different view model instance for different parameters.

Would it make sense to treat `parameters` just like the `key` parameter?
Otherwise we'd need to always pass our parameter to `key` redundantly.
In case of multiple parameters, we'd need to map them to a single value to pass them to the `key` parameter.

Are there case, where parameters shouldn't create a different view model instance?

snaum47:
[...]

Issue #1477, opened 2022-11-17
Compose - getViewModel with new parameters doesn't give a new instance - androidx-compose:3.2.2


oteljazouli-bc:
**Describe the bug**
```
userId?.let {
    ActivityDetail(
        viewModel = getViewModel(
            scope = myScope,
            parameters = {
                println("userId $userId")
                parametersOf(userId) }
        )
    )
} 
```

```
@Composable
fun ActivityDetail(
    viewModel: DetailViewModel
) {
```

I have this setup, when `ActivityDetail(` is called the first time, it gives the right instance with the right `userId`, closing that and calling again with another `userId` keep the same old instance of the viewModel with the old `userId`

I saw from https://github.com/InsertKoinIO/koin/issues/1305 that there was a similar fix for `3.2.1`, I used `3.2.1` and `3.2.2` but still having this issue.

**To Reproduce**
please see the code above

**Expected behavior**
A new viewmodel instance when parameters are different

**Koin project used and used version (please complete the following information):**
"io.insert-koin:koin-androidx-compose:3.2.2"

**Additional moduleDefinition**
Add any other moduleDefinition about the problem here.

Nek-12:
I'm also facing this issue with any injected dependency, not just the viewModel one.

p-fischer:
I'm also facing this issue with a parameterized view model (using the `getViewModel(parameters = { parametersOf(elementId) })` syntax).

If I'm navigating back (Jetpack Compose via `NavController`) and re-enter the screen, the view model is re-created with the correct parameters.

In my scenario, I stay within a single route and only open a dialog, which has its own view model as a parameter.
These dialogs can be opened by clicking on list elements - so the dialog composable and view model get passed different arguments values each time.

Is there a version of Koin, which creates new view model instances?

arnaudgiuliani:
did you to upgrade to androidx-compose 3.4.1?

lammertw:
I'm having the same issue. Version 3.4.1 does not seem to fix it.

snaum47:
We're having the same issue but not with Compose.
Inside our fragment we inject viewmodels with different parameters provided each time. 
What we expect is that since the parameters are different, a new instance should be provided. But that's not the case and the 1st instance with the 1st parameters is provided on all future injections of the viewmodel in the same fragment.

A basic case with the issue is shown below:
`
// the module
fun appModule() = module {
  viewModel { parameters ->
    TestViewModel(parameters.get())
  }
}
[...]

PR #1592, opened 2023-05-23
Don't ignore key to get a viewmodel when a qualifier is present

lammertw:
Right now, the key is ignored in case a qualifier is used. I don't think that's the correct behavior as it makes it now impossible to force the creation of a new view model in case parameters have changed that require a new view model.

Relates to https://github.com/InsertKoinIO/koin/issues/1578 and https://github.com/InsertKoinIO/koin/issues/1477

### DIFF
--- a/android/gradle/versions.gradle
+++ b/android/gradle/versions.gradle
@@ -1,6 +1,6 @@
 ext {
     // Koin Versions
-    koin_android_version = '3.4.0'
+    koin_android_version = '3.4.1-SNAPSHOT'
 
     // build Tools
     android_min_version = 14
--- a/core/gradle/versions.gradle
+++ b/core/gradle/versions.gradle
@@ -1,6 +1,6 @@
 ext {
     // Koin Versions
-    koin_version = '3.4.0'
+    koin_version = '3.4.1-SNAPSHOT'
     
     // Kotlin
     kotlin_version = '1.8.10'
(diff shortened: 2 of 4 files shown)


arnaudgiuliani:
Thanks for the proposal, the main thing here is that key and koin qualifier are overriding a bit. Do you have a clear usecase?

pedrofsn (approved):

lammertw:
Hi, any update on this? I see it has been approved, but not yet merged. I've really love to have this (as I'm currently using my own fork)

android/gradle/versions.gradle
```diff
@@ -1,6 +1,6 @@
 ext {
     // Koin Versions
-    koin_android_version = '3.4.0'
+    koin_android_version = '3.4.1-SNAPSHOT'
```
arnaudgiuliani, line 3: please don't touch version in your PR

core/gradle/versions.gradle
```diff
@@ -1,6 +1,6 @@
 ext {
     // Koin Versions
-    koin_version = '3.4.0'
+    koin_version = '3.4.1-SNAPSHOT'
```
arnaudgiuliani, line 3: please don't touch version in your PR

arnaudgiuliani:
there is conflicts, and need to revert version change as mentioned.

arnaudgiuliani:
I've rebased the branch on https://github.com/InsertKoinIO/koin/pull/1644

Issue #1505, opened 2023-01-17
The `scopedOf` extension function mismatches number of arguments greater than 10


gary0707:
Dear Team,

I am migrating one of my project to Koin 3.3.2. The project has one class that has 19 arguments in the constructor. It's possible to use the `factoryOf` extension function for it, but not the `scopedOf` method.

Looking at the sources of the `koin-core-jvm:3.3.2`:
```kotlin
inline fun <reified R, reified T1, reified T2, reified T3, reified T4, reified T5, reified T6, reified T7, reified T8, reified T9, reified T10, reified T11, reified T12, reified T13, reified T14, reified T15, reified T16, reified T17, reified T18, reified T19, reified T20, reified T21, reified T22> ScopeDSL.scopedOf(
    crossinline constructor: (T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11) -> R,
    noinline options: DefinitionOptions<R>? = null,
): KoinDefinition<R> = scoped { new(constructor) }.onOptions(options)
```

the number of reified parameters is not aligned with the number of the `crossinline constructro` lambda. 

Is it by design or is it a bug?

arnaudgiuliani:
fixed in 3.4.0 👍

gary0707:
Hi @arnaudgiuliani ,
I have just checked the v.3.4.0 but the issue is still there, this time for constructors having 13 args and more

PR #1621, opened 2023-07-14
Fix type parameter count for scopedOf methods

ErikVermunt-TomTom:
The `scopedOf` extension functions with more than 10 constructor parameters have an incorrect number of type parameters.
This causes compile time errors when using them: `Not enough information to infer type variable T12`

Fixes https://github.com/InsertKoinIO/koin/issues/1505


arnaudgiuliani:
Thanks for your contribution.

3.4.3 is already having it. I can close it, thanks 👍

PR #1632, opened 2023-08-06
Allow includes to take Collection

DebitCardz:
Allow for `includes` to take a `Collection` as its parameter.
Functionality the same as a List however, it allows for Sets and Lists to be used in conjunction.


arnaudgiuliani:
thanks

PR #1643, opened 2023-09-04
Fixing race condition in Scope - Fixed for 3.5.0

arnaudgiuliani:
Original PR from @octa-one 
--
A race condition may occur when using the same Scope from different threads.
Now only add and remove calls are synchronized. And operations between these calls can be in a race.

Imagine that 2 threads are simultaneously resolving instances from some scope. These instances need parameters to be created. This means that some of their dependencies will be found in `_parameterStack`.

We call the `resolveInstance` method simultaneously from both threads.
The first thread puts the parameters on the `_parameterStack`. The second thread does the same. Next, the first thread calls `_parameterStack.firstOrNull()` and gets parameters from the second thread.
It turns out that the threads swapped their parameters.

This could be solved by additional synchronization, but in fact the threads don't need to be synchronized here. Resolution of particular dependencies takes place on a single thread, the main thing is to make sure that the stack is not changed from the outside during the resolution.
ThreadLocal is suitable for this purpose.

And since we made `_parameterStack` thread local, we don't need to additionally synchronize access to it and can remove `KoinPlatformTools.synchronized(this)`.

### DIFF
--- a/core/koin-core/build.gradle
+++ b/core/koin-core/build.gradle
@@ -75,6 +75,7 @@ kotlin {
             dependsOn commonMain
             dependencies {
                 implementation 'org.jetbrains.kotlin:kotlin-stdlib-js'
+                implementation "co.touchlab:stately-concurrency:1.2.2"
             }
         }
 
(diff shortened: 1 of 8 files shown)


arnaudgiuliani:
following from https://github.com/InsertKoinIO/koin/pull/1561

arnaudgiuliani:
Code has been tested, with fixed tests.
Benchmark run also 👍

PR #1644, opened 2023-09-04
Dont ignore getviewmodel key

arnaudgiuliani:
Follow up on https://github.com/InsertKoinIO/koin/pull/1592 from @lammertw 

> Right now, the key is ignored in case a qualifier is used. I don't think that's the correct behavior as it makes it now impossible to force the creation of a new view model in case parameters have changed that require a new view model.

Relates to https://github.com/InsertKoinIO/koin/issues/1578 and https://github.com/InsertKoinIO/koin/issues/1477


snaum47:
One thing to note regarding https://github.com/InsertKoinIO/koin/issues/1578, these merged changes would not the fix the issue we are facing, mainly because using "by viewmodel" does not have an option to provide the key as a parameter. Can you provide any guidance on this?

PR #1648, opened 2023-09-07
[maintenance] androidContext redundant bind

GrzegorzBobryk:
`androidContext` is smart cast to `Application` in if condition. Therefore binding it again to application is redundant, it is the same type twice. Changed `binds` to `bind`.


arnaudgiuliani (approved):

arnaudgiuliani:
thanks

PR #1670, opened 2023-10-06
Documentation: Fix typo in docs/reference/koin-core/modules.md

gustavobarbosab:


pedrofsn (approved):

arnaudgiuliani (approved):

nachtien (approved):

docs/reference/koin-core/modules.md
```diff
@@ -253,20 +253,20 @@ startKoin { modules(featureModule1, featureModule2) }
 Notice that all modules will be included only once: `dataModule`, `domainModule`, `featureModule1`, `featureModule2`.
 
 :::info
-    If you have any compiling issue while including modules from the same file, either use `get()` Kotlin attribute operator on your module eitehr separate each moduel in files. See https://github.com/InsertKoinIO/koin/issues/1341 workaround
+    If you have any compiling issue while including modules from the same file, either use `get()` Kotlin attribute operator on your module either separate each module in files. See https://github.com/InsertKoinIO/koin/issues/1341 workaround
```
nachtien, line 256: This sentence reads kinda screwy to me. I think this makes sense and is what the author intended?
```suggestion
    If you have any compiling issue while including modules from the same file, use the `get()` Kotlin attribute operator on each module. See https://github.com/InsertKoinIO/koin/issues/1341 workaround
```

nachtien (changes requested):

docs/reference/koin-core/modules.md
gustavobarbosab: I believe that this note pertains to the creation of modules within the same Kotlin file. This practice can sometimes cause issues in compilation time, and the note intends to inform the developer about it. It recommends separating the modules into distinct files if this kind of problem is happening. At least it was what I understood.

Was this indeed your intention, @arnaudgiuliani?

docs/reference/koin-core/modules.md
```diff
@@ -99,7 +99,7 @@ This way, your share the definitions and avoid preallocate factories in a value.
 
 ## Overriding definition or module (before 3.1.0)
 
-Koin won't allow you to redefinition an already existing definition (type,name,path ...). You will an an error if you try this:
+Koin won't allow you to redefinition an already existing definition (type,name,path ...). You will an error if you try this:
```
kibettheophilus, line 102:
```suggestion
Koin won't allow you to redefine an already existing definition (type,name,path ...). You will get an error if you try this:
```

kibettheophilus: @gustavobarbosab some suggestions.

kibettheophilus (approved):

docs/reference/koin-core/modules.md
gustavobarbosab: Thanks @kibettheophilus, it makes sense.

kibettheophilus (approved):

nachtien (approved):

docs/reference/koin-core/modules.md
arnaudgiuliani: yes, lets keep to say we have both possibility to use get() or separate file. It's more a IDE/Kotlin compilation stuff problem with our `include` operator

PR #1692, opened 2023-11-06
Documentation: fixing spelling mistake and replacing reference to `multiplatform` to `ktor`

rory-instil:
Changing term in Ktor section of setup docs from `multiplatform` to `ktor`
Fixing spelling mistake in same section


arnaudgiuliani:
thanks 🙏

PR #1701, opened 2023-11-11
Add new KoinApplication Composable

Bloemcol:
## Context
After using the `KoinApplication` Composable in a different project, it felt weird to me that [you need a lambda to provide the modules to your Koin application](https://github.com/Bloemcol/koin/blob/add-another-koinapplication-composable/compose/koin-compose/src/commonMain/kotlin/org/koin/compose/KoinApplication.kt#L151).

This new Composable only requires a list of modules to work, making it just a bit easier to use.

#### Bonus change
Removed unused import from sample app.

### Note
**⚠️ Warning: light mode screenshots ahead!**

I did not use hyphens in the JavaDoc comment, as the docs are easier on the eye without them. Compare the JavaDoc hover preview in Android Studio of the KoinApplication Composable mentioned above
Screenshot 2023-11-11 at 09 03 20
with the new one.
Screenshot 2023-11-11 at 09 03 34

### DIFF
--- a/compose/koin-compose/src/commonMain/kotlin/org/koin/compose/KoinApplication.kt
+++ b/compose/koin-compose/src/commonMain/kotlin/org/koin/compose/KoinApplication.kt
@@ -117,6 +117,28 @@ fun KoinApplication(
 }
 
 /**
+ * Create a new Koin Application context for Compose.
+ *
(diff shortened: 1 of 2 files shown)


Bloemcol:
Please see #1704 for a PR with verified commits.

PR #1723, opened 2023-11-28
Fix remember usage when retrieving Koin context

arnaudgiuliani:


compose/sample-android-compose/src/main/java/org/koin/sample/androidx/compose/MainApplication.kt
```diff
@@ -13,7 +13,8 @@ class MainApplication : Application() {
         super.onCreate()
 
         startKoin {
-            androidLogger(Level.DEBUG)
+//            androidLogger(Level.DEBUG)
+            printLogger(Level.DEBUG)
```
Ariessanea1, line 17: Level buD

PR #1739, opened 2023-12-20
use io.github.gradle-nexus.publish-plugin

arnaudgiuliani:
use io.github.gradle-nexus.publish-plugin to help secure staging repositories. Would need to clean all this old gradle scripts ^^

thanks @martinbonnin 🙏


martinbonnin (approved):
🎉

Issue #1747, opened 2023-12-30
missing watchosDeviceArm64 target


suau:
**Describe the bug**
Since XCode14 the `ARCHS_STANDARD` for watchOS includes arm64. KMP already has added the target watchosDeviceArm64, but koin hasn't included it yet

**To Reproduce**
1. Create a KMP project
2. Add Koin
3. Add a watchOS app to the ios app
4. Add watchosArm64(), watchosDeviceArm64(), watchosSimulatorArm64() to the shared build.gradle.kts
5. Build for release
6. Release build will fail with default settings

**Expected behavior**
watchos should build with default settings

**Koin module and version:**
`koin-core:3.5.3`

arnaudgiuliani:
Thanks, let's continue with the PR 👍

PR #1751, opened 2024-01-04
Add watchos support

suau:
Resolves watchos compatibility issue #1747 by adding `watchosDeviceArm64` target available since KMP 1.8.20


arnaudgiuliani:
let's go 👍

PR #1756, opened 2024-01-11
Add Lazy<Module> concrete implementation & missing "plus" operator

### LINKED ISSUE #1755: Missing "plus" operator for Lazy<Module>
The `Module` class implements the "plus" operator which can be widely used in a codebase. Given a migration path to lazy APIs (meaning migrating from using `module {}` to `lazyModule {}`) there is no equivalent API for using "plus" operator.

A "plus" operator for `Lazy<Module>` similar to what `Module` has.

Given a migration path there are no well-suited alternatives. Parity of operators should exist between `Lazy<Module>` and `Module`.

**Target Koin project**
`core`

## Note - Broken migration path
Considering the current implementation of the lazy modules, regardless of where the operator is implemented the migration path of a consumer from non-lazy to lazy APIs will require the addition of an import statement even though the "plus" operator of a `Module` does not (since the operator is implemented as a member of the class and not as an extension function). This is a side effect of not having a class for a "lazy module".

My suggestion as an overall improvement would be to create a `LazyModule` class that implements `Lazy<Module>` in order to:
1. Be compatible with the current APIs
2. Leverage delegation in order to implement `Lazy` logic
3. Having the option to keep parity between `Module` and `LazyModule` implementations

cpaleop:
Resolves #1755 by creating a `LazyModule` class that implements `Lazy<Module>` and also adds the missing "plus" (`+`) operator.


core/koin-core-coroutines/src/commonMain/kotlin/org/koin/core/module/LazyModule.kt
```diff
@@ -0,0 +1,45 @@
...
+ * Koin Lazy Module - an implementation of [Lazy]<[Module]>.
+ *
+ * Accepts a lambda that initializes a [Module] via [lazy()][lazy]
+ * using [LazyThreadSafetyMode.NONE] as thread-safety [mode][LazyThreadSafetyMode].
+ *
+ * @author Chris Paleopanos
+ * @param moduleInitializer a lambda that will be used to initialize a [Module] lazily
+ */
+@KoinDslMarker
+class LazyModule(moduleInitializer: () -> Module) : Lazy<Module> by lazy(LazyThreadSafetyMode.NONE, moduleInitializer) {
```
cpaleop, line 28: Let me know if you want me to annotate this with `@KoinExperimentalAPI`. I wasn't sure if that's desirable since that's not the behavior of the current implementation.

Annotating this class with `@KoinExperimentalAPI` will raise a warning for all modules that are constructed via `lazyModule {}` DSL.

arnaudgiuliani: this is already the case for LazyModule API no? 🤔

arnaudgiuliani: but I could raise it in 3.6

cpaleop: Currently the warning is raised on `lazyModule {}` usage. If I add the annotation here then the warning will also be raised for `lazyKoinModule` usages too in the following example - which is not the case so far because it's type is `Lazy<Module>`.

```kotlin
val lazyKoinModule = lazyModule {}
```

It makes sense to add the annotation here, but will increase the amount of warnings dramatically - up to you!

arnaudgiuliani: yes that was a case in previous Koin 2.x. Let's keep out of annotation then

arnaudgiuliani (approved):
Let's play with more test sample if we can :)

cpaleop:
Resolved the conflicts, let me know if there is anything else I need to care of :)

arnaudgiuliani:
Thanks, should be ok.

PR #1763, opened 2024-01-19
Improving Development Support For Android Jetpack Compose Previews

YanneckReiss:
### Problem description this pull request aims to resolve
When we have a single composable in Android Jetpack Compose that depends on injecting some definitions, we can easily use the `KoinApplication` composable, like currently suggested in the [Koin documentation for Compose Previews](https://insert-koin.io/docs/reference/koin-compose/compose#compose-preview-with-koin).

However, it's often the case that we want to define multiple compose previews with different use cases states of our composable. Unfortunately, when we have two or more compose previews in a single file, using the `KoinApplication` composable in both of them leads to a `org.koin.core.error.KoinAppAlreadyStartedException` because the first started `KoinApplication` also lives in the second compose preview. 

The problem here is that we often want to provide a different set of definitions (a different list of modules), or simply can't guarantee which of the compose previews will build first, and therefore don't know if we need to start another `KoinApplication` or just can use the `KoinContext` composable to provide a `KoinContext`. Another solution would be to manually create an isolated context for each of the compose previews and supply it to a `KoinApplication` , which would introduce a lot of boilerplate code. 

To showcase this problem at a practical example I [created a small sample repository](https://github.com/YanneckReiss/showcasebrokenkoincomposepreview).

### Description of my proposal
My solution for this problem is contained in this pull request. It should make the life easier for all developers who want to create compose previews for composables that rely on Koin.

If we come back to the mentioned example repository and take a look at the `ExampleView` instead of the following which throws the `org.koin.core.error.KoinAppAlreadyStartedException`:

```
@Preview
@Composable
fun Preview_ExampleView_enabled() {
    KoinApplication(application = { modules(appModule) }) {
        AppTheme {
            ExampleView(
                isEnabled = true
            )
        }
    }
}

@Preview
@Composable
fun Preview_ExampleView_disabled() {
    KoinApplication(application = { modules(appModule) }) {
        AppTheme {
            ExampleView(
                isEnabled = false
            )
        }
    }
}
```

we can simply replace the code with the following and don't need to worry about managing a compose preview (file) wide Koin context:

```
@Preview
@Composable
fun Preview_ExampleView_enabled() {
    KoinPreviewApplication(modules = { listOf(appModule) }) {
        AppTheme {
            ExampleView(
                isEnabled = true
            )
        }
    }
}

@Preview
@Composable
fun Preview_ExampleView_disabled() {
    KoinPreviewApplication(modules = { listOf(appModule) }) {
        AppTheme {
            ExampleView(
                isEnabled = false
            )
        }
    }
```

### DIFF
--- a/compose/koin-compose/src/commonMain/kotlin/org/koin/compose/preview/KoinPreviewApplication.kt
+++ b/compose/koin-compose/src/commonMain/kotlin/org/koin/compose/preview/KoinPreviewApplication.kt
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2017-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.koin.compose.preview
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import org.koin.compose.KoinIsolatedContext
+import org.koin.core.context.stopKoin
+import org.koin.core.module.Module
+import org.koin.core.KoinApplication
+import org.koin.dsl.koinApplication
+
+/**
+ * Starts a new isolated KoinApplication for the provided list of Koin modules.
+ * Can be used if you have multiple compose previews in one file that require certain definitions.
+ *
+ * @param modules list of definitions required for the preview
+ * @param content compose preview content
+ *
+ * @author Yanneck Reiß
+ */
+@Composable
+fun KoinPreviewApplication(
+    modules: () -> List<Module>,
+    content: @Composable () -> Unit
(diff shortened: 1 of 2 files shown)


YanneckReiss:
I just realized a drawback with this approach. It only works for constructor injected parameters. When one uses the `KoinComponent` for field injection and the respective class needs access to that dependency in the context of that preview (for example at class initialization) the isolated context we defined in the preview wouldn't apply because in the KoinComponent we rely on the Koin `GlobalContext` .. therefore using field injection with this approach leads to another exception:

`java.lang.IllegalStateException: KoinApplication has not been started
	at org.koin.core.context.GlobalContext.get(GlobalContext.kt:36)
	`
	
To visualize it, the following example with constructor injection works:

```
class TestClass(
    private val myInjectedClass: MyInjectedClass
)
```

While this example won't work because we eagerly inject the MyInjectedClass and therefore access the `GlobalContext` under the hood:
```
class TestClass: KoinComponent {

    private val myInjectedClass: MyInjectedClass = get()
}
```	

The [Koin documentation](https://insert-koin.io/docs/reference/koin-core/context-isolation/) suggests to define a custom `KoinComponent` but that wouldn't work for this case because we dynamically create a custom context for each preview. @arnaudgiuliani therefore maybe this pull request is obsolete or maybe you have an idea how to workaround this problem?

arnaudgiuliani:
we may need to investigate more to help cover the general case, and see the edge cases around 🤔 
Let's keep it as a base of suggestion. You can also restart a PR if you want

arnaudgiuliani:
can be moved into koin-compose-viewmodel

arnaudgiuliani:
Otherwise idea sounds good. We need to tag it Experimental to get feedback

arnaudgiuliani:
Just went with https://github.com/InsertKoinIO/koin/pull/2182
Can you give some feedback? Let's reopen something around, to keep things simple.

YanneckReiss:
The #2182 looks very promising @arnaudgiuliani !

Issue #1771, opened 2024-01-29
Implicit type resolution


peter-kovacs-accedo:
In my project when I want to use `get()` or `inject()` functions I have to specify types explicitly, which makes huge overhead in the project.

**See PR: https://github.com/InsertKoinIO/koin/pull/1772**

For more context: I'm working on a feature based application, where lots of interfaces can be injected with different implementations (only one by each sub-project) and it's handled when a dependency is not present on a given sub-project.
When I want to use `inject()` or `get()` I'm getting errors like: 

> No value passed for parameter 'clazz'

> Type mismatch (Required:Context Found: Lazy<???>)

> None of the following candidates is applicable because of receiver type mismatch
etc.

However it's simple to solve.

arnaudgiuliani:
Is it Java compat extension?

peter-kovacs-accedo:
No, I'm using those for kotlin. The issue is that Koin cannot resolve types which are coming from other packages.

Before this update I needed to use
`protected val myService: MyService by inject(MyService::class.java)`

With the changes it's enough to call:
`protected val myService: MyService by inject()'

peter-kovacs-accedo:
Just checking in, Did you evaluate this PR?

arnaudgiuliani:
what's the interest here? Java can't go over reified type, as a backward compact API.
You are using directly inject and get, where you should use either KoinComponent or current available extension. Also prefer go with constructor injection

muzzah:
I know im doing something wrong but Im getting the same error here in a kotlin codebase. Just started using Koin but cannot seem to get past this issue. Any help would be appreciated. 

using `io.insert-koin:koin-core:4.0.2`

<img width="1070" alt="Image" src="https://github.com/user-attachments/assets/7f0a5caa-4d34-4183-af57-b2e36d0719a0" />

PR #1772, opened 2024-01-29
Implicit type resolution for inject and get methods

peter-kovacs-accedo:
PR for issue: https://github.com/InsertKoinIO/koin/issues/1771


arnaudgiuliani:
what's the interest here? Java can't go over reified type, as a backward compact API.
You are using directly inject and get, where you should use either KoinComponent or current available extension. Also prefer go with constructor injection

arnaudgiuliani:
let's continue over the issue

PR #1777, opened 2024-02-02
[maintenance] bom added to versions table

GrzegorzBobryk:
Documentation is missing on bom version. 
Done in separate commits:
- row with bom version added to the table
- table reformatted


arnaudgiuliani:
thanks

Issue #1785, opened 2024-02-15
`stately-common` transitive dependency causing issues with other dependencies


kpgalligan:
**Describe the bug**
See [sqldelight bug](https://github.com/cashapp/sqldelight/issues/4357) for details.

**To Reproduce**
See [sqldelight bug](https://github.com/cashapp/sqldelight/issues/4357) for details.

**Expected behavior**
It doesn't do the thing described in the [sqldelight bug](https://github.com/cashapp/sqldelight/issues/4357).

**Koin module and version:**
Kotlin/Native targets, latest version(s)

**Snippet or Sample project to help reproduce**
I'm going to submit a PR. This is a tracking issue.

PR #1786, opened 2024-02-15
Bump version

kpgalligan:
PR for https://github.com/InsertKoinIO/koin/issues/1785. Will let CI run to determine if everything is OK, but it should be fine.


arnaudgiuliani:
thanks 👍 I've seen someone suggesting to go 2.0.6. I'll add that quickly

arnaudgiuliani (approved):

kpgalligan:
@arnaudgiuliani It might actually be easier if you make the change if it's just the version bump, as I need to figure out signing of commits, but my day is pretty tight. However, there's some kind of test failure, so I may need to dive into that anyway. It is a concurrency test failure, so it seems very likely related somehow (although there really shouldn't be any functional change with version 2.0.6, so I'm not sure *why* there would be an issue.

arnaudgiuliani:
no worry ... macos test failures are there since a while 🙄

Issue #1792, opened 2024-02-23
Use ktor-koin3.5.2-RC1, with ktor auto-reload, not working ktor-server.


momosetkn:
**Describe the bug**

becomes unresponsive after ktor auto-reload.

**To Reproduce**
1. git clone https://github.com/momosetkn/koin-ktor-auto-reload-bug
2. `./gradlew build and `./gradlew -t compilekotlin`
3. execute below command in another console. `./gradlew run`
4. when you change the code, It becomes unresponsive, like below.
```
curl localhost:8080
# curl: (52) Empty reply from server
```

**Expected behavior**
server was respond

**Koin module and version:**

ktor-koin version: 3.5.2-RC1
(3.5.1 version is works)

**Snippet or Sample project to help reproduce**

https://github.com/momosetkn/koin-ktor-auto-reload-bug

arnaudgiuliani:
Follow up PR has been merged 👍

Issue #1782, opened 2024-02-12
Koin+Ktor: first server reload breaks consequent server responses


Stexxe:
When the Ktor's auto-reloading feature is enabled, and the `Koin` Ktor plugin is installed, the server doesn't respond after the first reload.

**To Reproduce**
1. Run the server with the following code and the [auto-reloading enabled](https://ktor.io/docs/auto-reload.html#enable):
```kotlin
fun main() {
    embeddedServer(Netty, port = 8080) {
        install(Koin)
        routing {
            get("/") {
                call.respondText("Hello, world")
            }
        }
    }.start(wait = true)
}
```
2. Make a request to `http://localhost:8080`
3. Change the string in `call.respondText("Hello, world")`
4. Build the project
5. Make the request again

**Expected behavior:**
A response with the body is returned from the server
**Actual behavior:**
An empty response is returned from the server. Also, I observe the following exception being thrown:
```
org.koin.core.error.KoinAppAlreadyStartedException: A Koin Application has already been started
	at org.koin.core.context.GlobalContext.register(GlobalContext.kt:44)
	at org.koin.core.context.GlobalContext.startKoin(GlobalContext.kt:56)
	at org.koin.core.context.DefaultContextExtKt.startKoin(DefaultContextExt.kt:34)
	at org.koin.ktor.plugin.KoinPluginKt$Koin$2.invoke(KoinPlugin.kt:37)
	at org.koin.ktor.plugin.KoinPluginKt$Koin$2.invoke(KoinPlugin.kt:35)
	at io.ktor.server.application.CreatePluginUtilsKt.setupPlugin(CreatePluginUtils.kt:301)
	at io.ktor.server.application.CreatePluginUtilsKt.createPluginInstance(CreatePluginUtils.kt:270)
	at io.ktor.server.application.CreatePluginUtilsKt.access$createPluginInstance(CreatePluginUtils.kt:1)
	at io.ktor.server.application.CreatePluginUtilsKt$createApplicationPlugin$2.install(CreatePluginUtils.kt:90)
	at io.ktor.server.application.CreatePluginUtilsKt$createApplicationPlugin$2.install(CreatePluginUtils.kt:83)
	at io.ktor.server.application.ApplicationPluginKt.install(ApplicationPlugin.kt:100)
	at io.ktor.server.application.ApplicationPluginKt.install$default(ApplicationPlugin.kt:88)
	at ServerKt$main$1.invoke(Server.kt:10)
[...]

PR #1797, opened 2024-02-25
Fix for can auto-reload on ktor

momosetkn:
Fix not working ktor-server auto-reload problem(#1792, #1782).


projects/ktor/koin-ktor/src/main/kotlin/org/koin/ktor/plugin/KoinPlugin.kt
```diff
@@ -34,7 +35,8 @@ import org.koin.dsl.KoinAppDeclaration
  */
 val Koin = createApplicationPlugin(name = "Koin", createConfiguration = { KoinApplication.init() }) {
     val koinApplication = setupKoinApplication()
-    startKoin(koinApplication)
+    runningKoinApplication?.let { stopKoin() } // for ktor auto-reload
```
arnaudgiuliani, line 38: You are keeping a ref on the Koin context instance here. Either I would suggest using `KoinPlatformTools.defaultContext().getOrNull()` to request current Koin context and see if it's null

arnaudgiuliani:
added for 3.5.4 fixes

Issue #1794, opened 2024-02-24
[koin-core] flatten function


pedrofsn:
I was analyzing an app initialization and I saw Koin consuming a lot of time in _flatten_ function.
**This logic could be improved?**

Flatten function:
```
@OptIn(KoinInternalApi::class)
fun flatten(modules: List<Module>): Set<Module> {
    fun flat(modules: List<Module>, newModules: MutableSet<Module>){
        modules.forEach{
            newModules += it
            flat(it.includedModules,newModules)
        }
    }
    return mutableSetOf<Module>().apply { flat(modules,this) }
}
```
I was thinking in process this in IO instead of Main thread. In this case Koin will need to open this to the client pass the flattened list.

Another idea is try to figure out if it possible to apply other algorithm to reduce the time-space complexity.

pedrofsn:
Chart generated with macrobenchmark library, with Koin consuming a lot.
![chart](https://github.com/InsertKoinIO/koin/assets/3484814/45778a10-2dcb-41cb-94d0-859d1769ed6b)

pedrofsn:
We did a POC moving our koin to the background, using lazyLoadModules. And it is charming, but don't solve the problem with _flatten_ function.
![image](https://github.com/InsertKoinIO/koin/assets/3484814/bb6c8e60-76c5-42a4-8292-0ccda18b5831)

pedrofsn:
Thank you @tqi-gmagro for providing this analysis and collecting the charts!

arnaudgiuliani:
there is a first proposal with #1801 - let's try to benchmark results

arnaudgiuliani:
see #1801

PR #1801, opened 2024-03-01
flatten: use loop and ArrayDeque instead of recursion

hoc081098:
maybe fix https://github.com/InsertKoinIO/koin/issues/1794


arnaudgiuliani:
Could run some benchmark test to try to ensure we don't reproduce the same @hoc081098 ? 🙏

hoc081098:
> Could run some benchmark test to try to ensure we don't reproduce the same @hoc081098 ? 🙏 

Yes, I will add it 😊

hoc081098:
@arnaudgiuliani I have added some benchmarks.
```
BenchmarkClass.flatten            avgt   10  146.838 ± 15.775  ms/op
BenchmarkClass.newFlatten         avgt   10    0.880 ±  0.018  ms/op
```

- `BenchmarkClass.flatten` uses flatten of current Koin 3.5.3.
- `BenchmarkClass.newFlatten` uses the new implementation of flatten.

Could you check it again 🙏 ?

examples/jvm-perfs/benchmark.txt
marcellogalhardo, line 1: These are very nice numbers, thank you for taking the time!

I would avoid using `new` prefix, as that may change in the future. Maybe we can call them:
- flatten_recursive (or tailrec)
- flatten_ iterative

examples/jvm-perfs/src/main/kotlin/org/koin/benchmark/nestedModules.kt
marcellogalhardo, line 1: I like how you dynamically create the modules but I think the most important performance aspect here is the **depth** of the nested modules.

What if we dynamically create a N modules with N of depth, using a loop which we create a dynamic module and return itself, and include that in the "outer level" instead of a fixed depth of 4?

Opt: I would also consider to add more than one threshold to see how the latency increase with the number of modules (maybe 100, 1000 and 100000 if that is doable)?

examples/jvm-perfs/src/main/kotlin/org/koin/benchmark/nestedModules.kt
```diff
@@ -0,0 +1,99 @@
...
+        single(q) { A16() }
+        single(q) { A17() }
+        single(q) { A18() }
+        single(q) { A19() }
+        single(q) { A20() }
+    }
+}
+
+@OptIn(KoinInternalApi::class)
+internal fun newFlatten(modules: List<Module>): Set<Module> {
```
marcellogalhardo, line 84: Out of curiosity: can't you use the original function here instead of duplicating it?

projects/core/koin-core/src/commonMain/kotlin/org/koin/core/module/Module.kt
```diff
@@ -235,12 +235,19 @@ operator fun List<Module>.plus(module: Module): List<Module> = this + listOf(mod
  * Run through the module list to flatten all modules & submodules
  */
 @OptIn(KoinInternalApi::class)
-fun flatten(modules: List<Module>): Set<Module> {
-    fun flat(modules: List<Module>, newModules: MutableSet<Module>){
-        modules.forEach{
-            newModules += it
-            flat(it.includedModules,newModules)
+fun flatten(modules: List<Module>): Set<Module> {
```
marcellogalhardo, line 238: No action needed: that is nice and it is very similar to the original `flatten` function I wrote when introducing the `includes` feature [here](https://github.com/InsertKoinIO/koin/pull/1210/files#diff-e772f920a0a8f0cef44eb951fbb66027f8ee3b191a8cd91bc25b5f188ffaf837R338-R357) (for context).

projects/core/koin-core/src/commonMain/kotlin/org/koin/core/module/Module.kt
```diff
@@ -235,12 +235,19 @@ operator fun List<Module>.plus(module: Module): List<Module> = this + listOf(mod
...
-            newModules += it
-            flat(it.includedModules,newModules)
+fun flatten(modules: List<Module>): Set<Module> {    
+    val flatten = HashSet<Module>()
+    val stack = ArrayDeque<Module>(modules)
+
+    while (stack.isNotEmpty()) {
+        val current = stack.removeLast()
+        flatten.add(current)
+        current.includedModules.forEach {
```
marcellogalhardo, line 245: Please use a `for` loop instead, `forEach` is designed to be used in "chains" or if the receiver is nullable.

For a reference, please see the [Kotlin Style Guide](https://kotlinlang.org/docs/coding-conventions.html#loops):
> `forEach` (prefer using a regular for loop instead, unless the receiver of `forEach` is nullable or `forEach` is used as part of a longer call chain.

projects/core/koin-core/src/commonMain/kotlin/org/koin/core/module/Module.kt
```diff
@@ -235,12 +235,19 @@ operator fun List<Module>.plus(module: Module): List<Module> = this + listOf(mod
...
+            if (it !in flatten) {
```
marcellogalhardo, line 246: Nit: use a descriptive name for `it`.

marcellogalhardo (approved):
I was planning to take on this issue, but you were faster! Your implementation is exactly what I had in mind, so thank you for contributing with the improvement.

arnaudgiuliani:
Good work. I need to redirect it on 3.6 branch ... will do a merge from main to 3.6

arnaudgiuliani:
macos tests are fixed in another branch

examples/jvm-perfs/src/main/kotlin/org/koin/benchmark/nestedModules.kt
```diff
@@ -0,0 +1,99 @@
...
+    val flatten = HashSet<Module>()
+    val stack = ArrayDeque<Module>(modules)
+
+    while (stack.isNotEmpty()) {
+        val current = stack.removeLast()
+        flatten.add(current)
```
marcellogalhardo, line 90: Nit: `+=`

examples/jvm-perfs/src/main/kotlin/org/koin/benchmark/nestedModules.kt
```diff
@@ -0,0 +1,99 @@
...
+        current.includedModules.forEach {
+            if (it !in flatten) {
+                stack.addLast(it)
```
marcellogalhardo, line 93: Nit: `+=` (fyi: it uses `addLast` behind the scenes).

hoc081098:
@arnaudgiuliani Should I change target branch to 3.6?

hoangchungk53qx1:
:pray

projects/core/koin-core/src/commonMain/kotlin/org/koin/core/module/Module.kt
```diff
@@ -236,11 +236,27 @@ operator fun List<Module>.plus(module: Module): List<Module> = this + listOf(mod
  */
 @OptIn(KoinInternalApi::class)
 fun flatten(modules: List<Module>): Set<Module> {
-    fun flat(modules: List<Module>, newModules: MutableSet<Module>){
-        modules.forEach{
-            newModules += it
-            flat(it.includedModules,newModules)
+    // This is essentially a DFS traversal of the module graph,
```
hoc081098, line 239:
```suggestion
    // This is actually a DFS traversal of the module graph,
```

examples/jvm-perfs/src/main/kotlin/org/koin/benchmark/nestedModules.kt
```diff
@@ -0,0 +1,58 @@
...
+            }
+        }
+    }
+
+    return matrix.first()
+}
+
+@OptIn(KoinInternalApi::class)
+internal fun flattenIterative(modules: List<Module>): Set<Module> {
+    // This is essentially a DFS traversal of the module graph,
```
hoc081098, line 35:
```suggestion
    // This is actually a DFS traversal of the module graph,
```

hoc081098:
@arnaudgiuliani please review it again 🙏

arnaudgiuliani:
@hoc081098 The optimization is really great but we have an issue as, the returned result can randomly not be in the same order as the module list. 

flatten(m1, m2) can result in <m1,m2> or <m1,m2> which is breaking capacity to overload a definition, given the module order

arnaudgiuliani:
`DefinitionOverrideTest` is failing on branch 3.6.0

hoc081098:
> `DefinitionOverrideTest` is failing on branch 3.6.0

I think because I have used HashMap, I will create a PR to fix it

hoc081098:
> `DefinitionOverrideTest` is failing on branch 3.6.0

@arnaudgiuliani 
I think because I have used HashSet, I will create a PR to fix it

hoangchungk53qx1:
I think so, linkedhashmap is better than not.

Vào 9:22 Th 6, 5 thg 4, 2024 Petrus Nguyễn Thái Học <
***@***.***> đã viết:

hoc081098:
@arnaudgiuliani please review PR #1841.
Many thanks 🙇

PR #1819, opened 2024-03-11
Fix #1818 - Route based navigation support

### LINKED ISSUE #1818: Add route support to koinNavGraphViewModel() extension
The koinNavGraphViewModel() extension is perfect for scoping ViewModels to nav graphs but when using the [nav graph DSL](https://developer.android.com/guide/navigation/design/kotlin-dsl) graph ids are inaccessible.

It would be nice if the extension function also supports routes. This issue was reported earlier as https://github.com/InsertKoinIO/koin/issues/1476 and addressed in this PR https://github.com/InsertKoinIO/koin/pull/1473 but the issue and PR were closed without being merged in.

Currently I use my own variant of the extension function which does support route based navigation. It would however be better if I could just use the koinNavGraphViewModel() extension function.

**Target Koin project**
koin-androidx-navigation

SebastianHelzer:
Adds missing support to the `koin-androidx-navigation` library for route based navigation so that users of the [Kotlin DSL](https://developer.android.com/guide/navigation/design/kotlin-dsl) for the [Navigation library](https://developer.android.com/jetpack/androidx/releases/navigation) can have nav graph scoped view models.


projects/android/koin-androidx-navigation/src/main/java/org/koin/androidx/navigation/NavGraphExt.kt
```diff
@@ -41,3 +41,20 @@ inline fun <reified VM : ViewModel> Fragment.koinNavGraphViewModel(
...
+
+/**
+ * Request a ViewModel instance, scoped to Navigation graph
+ *
+ * @param route
+ *
+ * @author Sebastian Helzer
+ */
+inline fun <reified VM : ViewModel> Fragment.koinNavGraphViewModel(
+    route: Int,
```
hijano87, line 53:
```suggestion
    route: String,
```

Issue #1832, opened 2024-03-27
Compose view model constantly returning new instances with non-root scope


thadcodes:
**Describe the bug**
If using compose you fetch two different view model classes using `org.koin.androidx.compose.koinViewModel` with a non-root scope, every time a recompose happens a new instance of each view model will be returned.  

*Note*: I know the issue is with the way the key is being generated and will submit a PR shortly as well

**To Reproduce**
1. Run Sample application
2. Read the logcat entries after a moment of displaying

**Expected behavior**
Every time `view model: {Instance}" is printed it should be the same time

**Koin module and version:**
`koin-androidx-compose:3.5.3`
`koin-android:3.5.3`

**Snippet or Sample project to help reproduce**
Sample project to reproduce the issue.  It's a reused project from reporting another issue so disregard the name
[koinQualifierIssue.zip](https://github.com/InsertKoinIO/koin/files/14775720/koinQualifierIssue.zip)

wax911:
@thadcodes out of curiosity, might this also be related to https://github.com/InsertKoinIO/koin/issues/1831 which sounds similar 🤔

thadcodes:
> @thadcodes out of curiosity, might this also be related to #1831 which sounds similar 🤔

Yes that could be the same, but part of what causes this is having 2 view model classes used in the same scope and same view lifecycle. I suspect a way to test is to give your view model a qualifier even if you don't need it and see if the issue goes away

arnaudgiuliani:
Yes, known issue in 3.5.3 due to VM key generation. It's fixed in 3.5.4.

thadcodes:
> Yes, known issue in 3.5.3 due to VM key generation. It's fixed in 3.5.4.

Can you point to the fix?  I'm unclear how this wouldn't still be happening on the main branch

PR #1833, opened 2024-03-27
Add model class to view model key when scope, qualifier, or key are provided to the view model fetch

thadcodes:
Fixes gh-1832


thadcodes:
This issue is fixed in 3.5.4 by removing scope id from the key, so withdrawn for now

PR #1841, opened 2024-04-05
fix(flatten): use `LinkedHashSet` instead of `HashSet`

hoc081098:
to maintain the insertion order

### DIFF
--- a/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/module/Module.kt
+++ b/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/module/Module.kt
@@ -239,7 +239,7 @@ fun flatten(modules: List<Module>): Set<Module> {
     // This is actually a DFS traversal of the module graph,
     // but we're using a stack instead of recursion to avoid stack overflows and performance overhead.
 
-    val flatten = HashSet<Module>()
+    val flatten = linkedSetOf<Module>()
     val stack = ArrayDeque(modules)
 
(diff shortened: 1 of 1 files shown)


arnaudgiuliani:
still get error, but order is maintained in the reverse order I see 🤔 

```kotlin
@Test
    fun `allow overrides by type`() {
        val app = koinApplication {
            modules(
                module {
                    single<Simple.ComponentInterface1> { Simple.Component2() }
                },
                module {
                    single<Simple.ComponentInterface1> { Simple.Component1() }
                },
            )
        }

        app.assertDefinitionsCount(1)
        assertTrue(app.koin.get<Simple.ComponentInterface1>() is Simple.Component1)
    }
```

Good thing is that order seems stable

image

arnaudgiuliani:
Could test that with this test too:

```kotlin
@Test
    fun test_flatten_list() {

        (1..100).all { i ->
            val modules = listOf(module { },module { },module { },module { },module { },module { },module { },module { },module { },module { })
            val flatten = flatten(modules)
            flatten.reversed() == modules
        }
    }
```

hoc081098:
@arnaudgiuliani just updated
```diff
-ArrayDeque(modules)
+ArrayDeque(modules.asReversed())
```

arnaudgiuliani:
Sounds good 🎉. Need to check the fails on test examples 🤔

arnaudgiuliani:
Seen the problem. It's not from your side 👍 Will fix it

PR #1846, opened 2024-04-09
Fix CI: update `examples/gradle/versions.gradle` kotlin to 1.9.23

hoc081098:
https://developer.android.com/jetpack/androidx/releases/compose-compiler#1.5.11


arnaudgiuliani:
yep 👍

PR #1854, opened 2024-04-13
Add manualVerification argument for test verification

JajaComp:
In my project, I run a dependency graph test from an application module, as a result of which I cannot access many internal module classes. This modification solves this problem. Also in the Android project there are many internal classes of the ClassNameArgs type, this modification also allows you to validate classes by mask.


arnaudgiuliani:
Hello, we need more details/test to illustrate your proposal

JajaComp:
I just want validate graph from root module in multi module project. I can't get access to internal class in sub modules

arnaudgiuliani:
Verification API now incoporate parameter injection description, and also generate any missing declaration - Koin 4.0

arnaudgiuliani:
Let's close this PR. Reopen something regarding Koin 4.0 if you need

PR #1865, opened 2024-05-02
Use explicit Gradle DependencyHandlerEnvorderPlatform

dragetd:
According to Jetbrains, using 'platform()' is deprecated and one should use the explicit 'project.dependencies.platform()' instead. See https://youtrack.jetbrains.com/issue/KT-58759/Deprecate-platform-enforcedPlatform-and-other-related-to-Gradle-DependencyHandler-methods-in-KotlinDependencyHandler

I do not really know much about the dependency handling logic. I only stumbled over this as I added Koin to my Compose project.


arnaudgiuliani:
thanks 👍

PR #1866, opened 2024-05-03
Add WasmJS support + fix test compilation for both JS and Wasm

JSMonk:


arnaudgiuliani:
Did you checked https://github.com/InsertKoinIO/koin/tree/3.6.0? 

It's already containing support for Wasm

Issue #1826, opened 2024-03-21
Support androidx viewmodel for kotlin multiplatform


Vivecstel:
https://developer.android.com/jetpack/androidx/releases/lifecycle#2.8.0-alpha03 added view model support for kotlin multiplatform.
Ideally koin should support this also.
Thanks

fluttertutorialin:
Can you explain how to add viewmodel library in kmm

marcellogalhardo:
An important information here is that AndroidX ViewModel does not include support for Web (js/wasm).

frankois1234:
Hello,
some reference here

- https://touchlab.co/kmp-viewmodel

A full demo here :
- https://github.com/joreilly/FantasyPremierLeague

It should be OK that koin move the viewmodel to KMP.

mdsadiqueinam:
The is the official repo used in google codelabe
https://github.com/MatkovIvan/nav_cupcake

AbdelrahmanEsam:
@frankois1234  it doesn't work with koin ....  iam getting this error 

Uncaught Kotlin exception: org.koin.core.error.InstanceCreationException: Could not create instance for '[Factory:'viewModel class Name']'
  at 0  shared

also I am creating this viewModel by factory or single ..... I think koin should support the viewModel{} at the common main

Leedwon:
@AbdelrahmanEsam 

Inspired by https://github.com/MatkovIvan/nav_cupcake/blob/master/composeApp/src/commonMain/kotlin/com/matkovivan/nav_cupcake/ViewModels.kt which uses androidx viewmodel. I'm using the following code to make it work with Koin:

`ViewModels.kt` in commonMain:
```Kotlin
internal expect fun <VM : ViewModel> viewModel(
    modelClass: KClass<VM>,
    factory: ViewModelProvider.Factory = rememberViewModelFactory()
): VM

// Can't be private as it causes Exception during IR lowering
@Composable
internal fun rememberViewModelFactory(): ViewModelProvider.Factory {
    val koin = getKoin()
    return remember {
        viewModelFactory {
            initializer { MyViewModel(myDependency = koin.get()) }
            ... // more VM initializers
        }
    }
}
```
`ViewModels.android.kt` in androidMain:
```Kotlin
@Composable
internal actual fun <VM : ViewModel> viewModel(
    modelClass: KClass<VM>,
    factory: ViewModelProvider.Factory
[...]

PR #1875, opened 2024-05-15
3.6.0 compose viewmodel - KMP

arnaudgiuliani:
- update latest compose 1.6.10
- lib deps with compose-viewmodel and navigation
- port of ViewModel API
- port of NavViewModel API

keep all @KoinExperimentalAPI for now

Fixing #1826


arnaudgiuliani:
iOS regression found due to https://github.com/JetBrains/compose-multiplatform/issues/4809

Dev is on hold then :/

arnaudgiuliani:
Should be good. Ported also `viewModel` and `viewModelOf` DSL to let define it in common code side.
See https://github.com/arnaudgiuliani/JetpackViewModel-CMP/blob/master/composeApp/src/commonMain/kotlin/AppModule.kt#L7

PR #1916, opened 2024-07-15
Update kmp.md

GabrielBrasileiro:
Update of KMP Native Memory Management reference.

When I reading the references I try to access the link but it's not with the correct reference.

I think the official page Kotlin/Native memory management﻿ updated at 21 May 2024 is the correctly.


arnaudgiuliani:
This section is not new anymore :P

Issue #1929, opened 2024-07-30
Koin(-core-viewmodel) 4.0.0-RC1 appears to depend on compose 1.7.0-alpha01


mikepenz:
**Describe the bug**

Koin 4.0.0-RC1 [contains a commit](https://github.com/InsertKoinIO/koin/commit/d38853b524ebc2cc32930e777466cd0c0d6642ac#diff-46bb88d929c23161f27e7968f73bd6f3a3e6c6f50041ff1c0f10b7072c661e3fR23) upgrading to compose-navigation 2.8.0-alpha08 - This change results in koin depending on compose 2.7.0-alpha01. 

Based on other version [definitions in the toml](https://github.com/InsertKoinIO/koin/commit/d38853b524ebc2cc32930e777466cd0c0d6642ac#diff-46bb88d929c23161f27e7968f73bd6f3a3e6c6f50041ff1c0f10b7072c661e3fR21) I believe the project means to still depend on compose 1.6.11 instead for the 4.0.0 release. 

Unfortunately, using `compose-navigation` 2.8.0-alpha08 forces the project to Compsoe 1.7.0 

<img width="1616" alt="Screenshot 2024-07-30 at 15 20 26" src="https://github.com/user-attachments/assets/9d08dd90-151a-40e8-bee6-8978e66dc90c">

While `compose-navigation` 2.7.0-alpha07 would use compose 1.6.11
 
<img width="1635" alt="Screenshot 2024-07-30 at 15 20 10" src="https://github.com/user-attachments/assets/e564f4e7-12cd-4d7c-a544-9b2dfa29188c">

This incompatibility results in projects built with compose 1.6.11 failing for example due to: 
```
Exception in thread "main" java.lang.NoSuchMethodError: 'void androidx.compose.runtime.Composer.startReplaceGroup(int)'
	at androidx.compose.ui.platform.PlatformLocalization_desktopKt.defaultPlatformLocalization(PlatformLocalization.desktop.kt)
```
(or other APIs which were changed like `ripple`, ..)

**To Reproduce**
1) Upgrade the project to Koin 4.0.0-RC01
2) Retrieve `dependencies` with gradles `dependencies` task
3) Observe compose 1.7.0-alpha01 is pulled in via the navigation plugin

```
+--- io.insert-koin:koin-compose-viewmodel -> 4.0.0-RC1
|    +--- io.insert-koin:koin-compose:4.0.0-RC1 (*)
[...]

PR #1953, opened 2024-08-27
Split VM / Nav API - Retrograde to composeNavigation = "2.8.0-alpha02"

arnaudgiuliani:
Split ViewModel to avoid leaks of Navigation API


arnaudgiuliani:
Fix https://github.com/InsertKoinIO/koin/issues/1929

PR #1927, opened 2024-07-29
Use kotlin.time API

pubiqq:
_This PR is a duplicate of https://github.com/InsertKoinIO/koin/pull/1697._

--

Kotlin Time API is now [stable](https://kotlinlang.org/docs/whatsnew19.html#stable-time-api), so it's time to replace the self-made time API with it.

`KoinPlatformTimeTools` and other parts of "koin time API" have been marked as deprecated and hopefully they will be removed in the future, as they shouldn't have been part of the public API.

PR #1955, opened 2024-08-27
Use kotlin.time API (for Koin 4.x)

pubiqq:
_This PR is the equivalent of https://github.com/InsertKoinIO/koin/pull/1927, but for the [`4.0.0`](https://github.com/InsertKoinIO/koin/tree/4.0.0) branch._

--

Kotlin Time API is now [stable](https://kotlinlang.org/docs/whatsnew19.html#stable-time-api), so it's time to replace the self-made time API with it.

`KoinPlatformTimeTools` and other parts of "koin time API" have been marked as deprecated and hopefully they will be removed in the future, as they shouldn't have been part of the public API.


arnaudgiuliani:
Already inside Koin 4.0. 👍 
Thanks for the feedback

arnaudgiuliani:
Reopen a PR if you see something interesting 👍

PR #1956, opened 2024-08-28
Fix koin-compose link on README.md

xxfast:


arnaudgiuliani:
No needed here anymore. transferred in setup/koin.md 👍

PR #1980, opened 2024-09-16
Fix formatting issue

SarahDelCastillo:
As it was written, the rendering in insert-koin.io/docs was wrong.


arnaudgiuliani:
thanks

PR #2020, opened 2024-10-10
[core] Ensure onClose callbacks are invoked in reversed initialization order

### LINKED ISSUE #1510: Order of onClose calls
### Discussed in https://github.com/InsertKoinIO/koin/discussions/1507

<div type='discussions-op-text'>

<sup>Originally posted by **smichovan** January 20, 2023</sup>
Is there a way how to change order in witch the onClose methods of components are called?

Lets have two services where the latter is dependent of the former: 

val appModule = module {
    single { Database() } onClose {it?.closeConnectionPool()} 
    single { ServiceUsingDb(get()) } onClose {it?.useDatabaseForCleanup()}
}

Database component maintains a connection to some DB and the second component accesses the DB through the first component. The "ServiceUsingDb" needs to have DB available in onClose method. As the "ServiceUsingDb" is actually dependent on "Database" I would expect Koin to ensure the right closing order. 

I made some research in the Koin code and if I understand it well, the components (instances, mappings) are stored in HashMap, means no specific order is guaranteed. In my case, it may (and does) happen that when ServiceUsingDb onClose is called, the DB is already closed and the cleanup fails. 

Have I missed anything? Is there any way how to influence the onClose methods order?
</div>

shchuko:
Fixes #1510

`onClose` callback can be used to release allocated resources
like DB connections. Discussion #1507

Some beans can depend on each other.

An example: classic Web application consists of `Controller`, `Service`, `Database`.

In this case, instantiation order is:
  1. `Database`
  2. `Service` - implements business logic, requires `Database`
  3. `Controller` - REST controller, requires `Service`

On application graceful shutdown, `onClose` invocation order must be reversed:
  1. `Controller` - created last, destroyed first
  2. `Service` - destroyed after `Controller` to avoid use-after-close
  3. `Database` - destroyed after `Service` to avoid use-after-close

This commit delivers implementation of `onClose` calls ordering.

When instance factory creates its first instance, it means that all dependencies are also created. So it's enough to sort factories by 'first instance creation,' and then reverse the sequence when invoking `onClose`.

Implementation:
- global 'instance creation order' atomic counter is created
- each instance factory has 'instance creation order' position
- when factory creates its first instance
  - 'instance creation order' counter is incremented
  - the counter's value is saved into 'instance creation order' position of the factory
- factories can be sorted by 'instance creation order' position, this can be used to order `onClose` calls

### DIFF
--- a/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/instance/InstanceFactory.kt
+++ b/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/instance/InstanceFactory.kt
@@ -17,6 +17,8 @@
 
 package org.koin.core.instance
 
+import co.touchlab.stately.concurrency.AtomicBoolean
+import co.touchlab.stately.concurrency.AtomicLong
 import org.koin.core.definition.BeanDefinition
 import org.koin.core.error.InstanceCreationException
 import org.koin.core.parameter.ParametersHolder
@@ -28,8 +30,12 @@ import org.koin.mp.Lockable
 /**
  * Koin Instance Holder
  * create/get/release an instance of given definition
+ *
+ * Implements [Comparable] to provide factories soring by instance creation order
  */
-abstract class InstanceFactory<T>(val beanDefinition: BeanDefinition<T>) : Lockable() {
+abstract class InstanceFactory<T>(val beanDefinition: BeanDefinition<T>) : Lockable(), Comparable<InstanceFactory<*>> {
+    private val instanceCreationOrderSet = AtomicBoolean(false)
+    private var instanceCreationOrderPosition = Long.MAX_VALUE
 
     /**
      * Retrieve an instance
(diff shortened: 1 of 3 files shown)


projects/core/koin-core/src/commonMain/kotlin/org/koin/core/instance/InstanceFactory.kt
```diff
@@ -28,8 +30,12 @@ import org.koin.mp.Lockable
 /**
  * Koin Instance Holder
  * create/get/release an instance of given definition
+ *
+ * Implements [Comparable] to provide factories soring by instance creation order
  */
-abstract class InstanceFactory<T>(val beanDefinition: BeanDefinition<T>) : Lockable() {
+abstract class InstanceFactory<T>(val beanDefinition: BeanDefinition<T>) : Lockable(), Comparable<InstanceFactory<*>> {
+    private val instanceCreationOrderSet = AtomicBoolean(false)
+    private var instanceCreationOrderPosition = Long.MAX_VALUE
```
marcellogalhardo, line 38: The current design ties a child instance _natural comparison_ to its insertion ordering within a parent collection. A `Comparable` is documented as a way to define a _class's natural order_, not their insertion order. See [Comparable][1] and [Natural Sort Order][2]

Instead of the child keeping track of the insertion order, what if the `InstanceRegistry` would handle the instances collection and ordering?

Here's a possible approach:

* The `InstanceRegistry` could use a `ConcurrentLinkedDeque` to store keys, ensuring thread-safe ordering of factory registrations.
 * The `InstanceRegistry` would then access the deque to order the factory instances map to close them in the correct order.

[1]: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Comparable.html
[2]: https://en.m.wikipedia.org/wiki/Natural_sort_order

arnaudgiuliani: same thought here. Consider another proposal @shchuko

arnaudgiuliani (changes requested):
Let's see for another solution 👍 See Marcello's comment

Issue #1963, opened 2024-09-03
ScopeRegistry.closeAllScopes() causes ConcurrentModificationException for JS


luozejiaqun:
Same issure to #1711 in JS platform. And the fix #1799 does't work for JS.
Here is the js `closeAllScopes` generated code:

```kotlin
private fun closeAllScopes() {
    _scopes.values.toList().forEach { scope ->
        scope.close()
    }
}
```

```javascript
function closeAllScopes($this) {
  // Inline function 'kotlin.collections.forEach' call
  var tmp0_iterator = $this._scopes_1.get_values_ksazhn_k$().iterator_jk1svi_k$();
  while (tmp0_iterator.hasNext_bitz1p_k$()) {
    var element = tmp0_iterator.next_20eer_k$();
    // Inline function 'org.koin.core.registry.ScopeRegistry.closeAllScopes.<anonymous>' call
    element.close_yn9xrc_k$();
  }
}
```

the method `toList` makes no sense to js.

luozejiaqun:
Instead of `toList`, maybe `mapTo` another list will help.

arnaudgiuliani:
I need to double-check with JetBrains team on this point, why it's producing such code

PR #2032, opened 2024-10-23
workaround ConcurrentModificationException when closeAllScopes in js

luozejiaqun:
Fixes gh-1963


arnaudgiuliani:
can you propose something else? it seems a bit weird here to do that 🤔

luozejiaqun:
@arnaudgiuliani just as i mentioned in #1963 , directly calling `scopes.toList()` on the JS platform is ineffective because it doesn't create a new list. Therefore, we need to explicitly create a separate list.

arnaudgiuliani:
@luozejiaqun we could test fix with

```kotlin
_scopes.values.toTypedArray().forEach { scope ->
            scope.close()
        }
```

arnaudgiuliani:
Let's see how 4.0.2-RC2 is going. Else let's reopen it 👍

PR #2081, opened 2024-12-17
Fix parameters stability inject

arnaudgiuliani:
- Fix ParametersHolder equals/hascode
- Fix koinInject function to stabilize new parameters

### DIFF
--- a/projects/compose/koin-compose/src/commonMain/kotlin/org/koin/compose/Inject.kt
+++ b/projects/compose/koin-compose/src/commonMain/kotlin/org/koin/compose/Inject.kt
@@ -16,11 +16,10 @@
 package org.koin.compose
 
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
-import androidx.compose.runtime.rememberUpdatedState
+import org.koin.core.annotation.KoinInternalApi
 import org.koin.core.parameter.ParametersDefinition
-import org.koin.core.parameter.emptyParametersHolder
+import org.koin.core.parameter.ParametersHolder
 import org.koin.core.qualifier.Qualifier
 import org.koin.core.scope.Scope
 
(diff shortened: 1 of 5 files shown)


projects/compose/koin-compose/src/commonMain/kotlin/org/koin/compose/Inject.kt
```diff
@@ -33,18 +32,15 @@ import org.koin.core.scope.Scope
...
     noinline parameters: ParametersDefinition? = null,
 ): T {
-    // This will always refer to the latest parameters
-    val currentParameters by rememberUpdatedState(parameters)
-
-    return remember(qualifier, scope) {
-        scope.get(qualifier) {
-            currentParameters?.invoke() ?: emptyParametersHolder()
-        }
+    val params: ParametersHolder? = parameters?.invoke()
```
Nek-12, line 42: This is incorrect as it will recreate parameters on every recomposition

projects/core/koin-core/src/commonMain/kotlin/org/koin/core/parameter/ParametersHolder.kt
```diff
@@ -139,6 +139,16 @@ open class ParametersHolder(
     }
 
     override fun toString(): String = "DefinitionParameters${_values.toList()}"
+
+    override fun equals(other: Any?): Boolean {
+        if (other is ParametersHolder){
```
Nek-12, line 144: This is incorrect as the equals check is unstable for parameter subclasses

projects/compose/koin-compose/src/commonMain/kotlin/org/koin/compose/Inject.kt
arnaudgiuliani: we don't have choice to have stable parametersHolder, else it's a function

PR #2082, opened 2024-12-17
Scope declared instance

arnaudgiuliani:
Replace usage of ScopedInstance holder to limit usage of declared instance to current scope, and avoid any reuse.

### DIFF
--- a/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/instance/DeclaredScopedInstance.kt
+++ b/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/instance/DeclaredScopedInstance.kt
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2017-Present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


projects/core/koin-core/src/commonMain/kotlin/org/koin/core/instance/DeclaredScopedInstance.kt
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Declared Instance in scope - Value holder to get back the value for the given scope Id
+ * to avoid ScopedInstanceFactory where we need the bean definition to return a definition
+ *
+ * @author Arnaud Giuliani
+ */
+class DeclaredScopedInstance<T>(beanDefinition: BeanDefinition<T>, val scopeID : ScopeID) :
+    InstanceFactory<T>(beanDefinition) {
+
+    private var value : T? = null
```
Nek-12, line 31: This is not thread-safe.

arnaudgiuliani: declare function is synchronized. But this will be fixed. I manage to fix that point

PR #1951, opened 2024-08-26
map & set multibinding

luozejiaqun:
Implement map & set multibinding just like dagger2.
Here is some implementation details:

Every MapMultibinding & SetMultibinding will involve 3 definitions:

1. the multibinding itself whose qualifier is mapMultibindingQualifier or setMultibindingQualifier
2. the element in multibinding whose qualifier is multibindingElementQualifier (use mapMultibindingQualifier/setMultibindingQualifier as prefix)
3. the MultibindingIterateKey which used to retrieve all elements, the key's qualifier is multibindingIterateKeyQualifier (also use mapMultibindingQualifier/setMultibindingQualifier as prefix)

When getting specific element from map multibinding, the map multibinding will try to retrieve the element through multibindingValueQualifier;
When getting all elements from map or set multibinding, the multibinding will first try to get all MultibindingIterateKeys and filter them based on multibindingQualifier, and then retrieve each element through multibindingElementQualifier.

The known problems:

1. the multibinding qualifier can't be null, since it's qualifier will be used as prefix of element & iterate key
2. the map multibinding doesn't support null as a key or value
3. ~~the elements order of multibiding is different with the order of element definitions~~ Now, both set & map multibinding elements are iterated in the order they are defined.

And just like dagger2, multibinding is inherited (because MultibindingIterateKey can be retrieved from getAll), current scope's multibinding can get all elements that define in linked scope.

Here are some use cases:

```kotlin
module {
    intoMap<String, Simple.ComponentInterface1>("keyOfComponent1") {
        Simple.Component1()
    }
    intoMap<String, Simple.ComponentInterface1>("keyOfComponent2") {
        Simple.Component2()
    }

    intoSet<Simple.ComponentInterface1> {
        Simple.Component1()
    }
    intoSet<Simple.ComponentInterface1> {
        Simple.Component2()
    }
}

val map: Map<String, Simple.ComponentInterface1> = koin.getMapMultibinding()
val set: Set<Simple.ComponentInterface1> = koin.getSetMultibinding()
```

issue #772

PR #2090, opened 2024-12-25
shouldn't inject parameters when qualifier is present

luozejiaqun:
example:
```kotlin
module {
    single(named("echo")) { "Hello," + it.get<String>() }
}

koin.get<String>(named("echo")) { parametersOf("World!") } // this should be "Hello,World!", but we get "World!"
```

PS. this issue will effect multibinding with parameters:
```kotlin
module {
    declareSetMultibinding<String> {
        intoSet { it.get<String>() + "1" }
        intoSet { it.get<String>() + "2" }
    }
}
koin.getSetMultibinding<String> { parametersOf("set element prefix") }
```
multibinding #1951

### DIFF
--- a/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/scope/Scope.kt
+++ b/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/scope/Scope.kt
@@ -322,7 +322,7 @@ class Scope(
     }
 
     private inline fun <T> resolveFromInjectedParameters(ctx: ResolutionContext): T? {
-        return if (ctx.parameters == null) null
+        return if (ctx.parameters == null || ctx.qualifier != null) null
             else {
             _koin.logger.debug("|- ? ${ctx.debugTag} look in injected parameters")
             ctx.parameters.getOrNull(clazz = ctx.clazz)
@@ -331,7 +331,7 @@ class Scope(
 
     private inline fun <T> resolveFromStackedParameters(ctx: ResolutionContext): T? {
         val current = parameterStack?.get()
-        return if (current.isNullOrEmpty()) null
+        return if (current.isNullOrEmpty() || ctx.qualifier != null) null
          else {
             _koin.logger.debug("|- ? ${ctx.debugTag} look in stack parameters")
             val parameters = current.firstOrNull()


luozejiaqun:
i also think it kind of weird that we can't do this:
```kotlin
module {
    single { "Hello," + it.get<String>() }
}

koin.get<String> { parametersOf("World!") } // never get "Hello,World!", always "World!"
```

when the declared type is same to the parameter and no qualifier presents, the parameter always win.

arnaudgiuliani:
I don't understand your point here 🤔 Let avoid mixing topic 👍

luozejiaqun:
The key point is the inject order:
```kotlin
// code from Scope
private fun <T> resolveFromContext(
  instanceContext: ResolutionContext
): T {
  return resolveFromInjectedParameters(instanceContext)
      ?: resolveFromRegistry(instanceContext)
      ?: resolveFromStackedParameters(instanceContext)
      ?: resolveFromScopeSource(instanceContext)
      ?: resolveFromParentScopes(instanceContext)
      ?: throwNoDefinitionFound(instanceContext)
}
```

If we want to resolve an instance of class `T`, and we provide an parameter which is also class `T`, then we will always get the parameter:

```kotlin
module {
    single<String> { parameters -> "Hello," + parameters.get<String>() }
}

koin.get<String> { parametersOf("World!") } == "World!"
```

This it the current implement, becase we can't tell which instance is wanted.
But, if we resolve an instance with qualifier, there is no way the parameter wins, since parameter has nothing to do with qualifier.

```kotlin
module {
    single<String>(named("echo")) { parameters -> "Hello," + parameters.get<String>() }
}

koin.get<String>(named("echo")) { parametersOf("World!") } // expected "Hello,World!", but currently we get "World!"
```

All in all, I just find a situation where we should skip the parameter injection and continue the rest.

arnaudgiuliani:
it's about how you shadow your definition instances. The latest additions in your context show be found first. If you try to pass an instance T in the parameters, you will solve this param first as it's the most recent additions.

If you use qualifier, it won't find it in the parameters as it doesn't have been tagged with qualifier.

luozejiaqun:
@arnaudgiuliani exactly, but the parameter injection doesn't check the qualifier, only the type:
```kotlin
    private inline fun <T> resolveFromInjectedParameters(ctx: ResolutionContext): T? {
        return if (ctx.parameters == null) null
            else {
            _koin.logger.debug("|- ? ${ctx.debugTag} look in injected parameters")
            ctx.parameters.getOrNull(clazz = ctx.clazz)
        }
    }
```

arnaudgiuliani:
I see your point, we could skip parameter stage if we have qualifier 👍

arnaudgiuliani:
New resolution engine is helping better guarding: https://github.com/InsertKoinIO/koin/blob/4.1.0/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/resolution/CoreResolver.kt#L58

Let me know if you see something else 👍

PR #2097, opened 2024-12-31
use inline value class to remove the runtime overhead of Qualifier

luozejiaqun:
`Qualifier` is ideally suited to be implemented as a value class. But the main problem is binary compatibility. The current implement is only source code compatible, not binary compatible. Maybe we should `internal` both `StringQualifier` & `TypeQualifier`, or deprecate them first.
This not a PR, but a suggestion to migrate `Qualifier` to value class.

### DIFF
--- a/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/qualifier/Qualifier.kt
+++ b/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/qualifier/Qualifier.kt
@@ -15,12 +15,15 @@
  */
 package org.koin.core.qualifier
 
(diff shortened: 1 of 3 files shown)


arnaudgiuliani:
@luozejiaqun can you rebase your PR? 🙏 
thanks

arnaudgiuliani:
Ok. Is it more an idea as hidding complexity behind Qualifier?

luozejiaqun:
@arnaudgiuliani I don't think so. While the qualifier itself is simple, there's complexity behind it. The goal is to simplify it. Since a qualifier is just a string, using a `value class` makes sense here and also removes runtime overhead.

PR #2105, opened 2025-01-03
check optional parameters during dependency verification

luozejiaqun:
Shouldn't throw `MissingKoinDefinitionException` if the parameter is optional.


arnaudgiuliani:
thanks

SarahDelCastillo:
Is there a specific configuration to respect in order for this to work ?
In my KMM project, I tried injecting a view model that has an optional parameter in its constructor, with the `viewModelOf` DSL, with `viewModel { ... }`, I also tried `factory { ... }` but I couldn't get it to work. I still have the `NoDefinitionFoundException` every time.

PR #2111, opened 2025-01-08
Fix DeclaredScopedInstance into adapted ScopedInstanceFactory

arnaudgiuliani:
Fix DeclaredScopedInstance solution to provide instance declared in rework previous with ScopedInstanceFactory

- design ScopedInstanceFactory to allow to not hold instance function, when declared on the fly
- added tests


projects/core/koin-core/src/commonMain/kotlin/org/koin/core/instance/ScopedInstanceFactory.kt
```diff
@@ -24,11 +25,18 @@ import org.koin.mp.KoinPlatformTools
...
+class ScopedInstanceFactory<T>(beanDefinition: BeanDefinition<T>, val holdInstance : Boolean = true) :
     InstanceFactory<T>(beanDefinition) {
 
     private var values = hashMapOf<ScopeID, T>()
 
+    fun size() = values.size
+
+    @PublishedApi
+    internal fun saveValue(id : ScopeID, value : T){
+        values[id] = value
```
Nek-12, line 37: This accessor is not thread safe

projects/core/koin-core/src/commonMain/kotlin/org/koin/core/instance/ScopedInstanceFactory.kt
```diff
@@ -42,20 +50,20 @@ class ScopedInstanceFactory<T>(beanDefinition: BeanDefinition<T>) :
...
         }
         KoinPlatformTools.synchronized(this) {
-            if (!isCreated(context)) {
-                values[context.scope.id] = create(context)
+            if (!isCreated(context) && holdInstance) {
+                values[context.scope.id] = super.create(context)
             }
         }
-        return values[context.scope.id] ?: error("Scoped instance not found for ${context.scope.id} in $beanDefinition")
+        return values[context.scope.id] ?: throw MissingScopeValueException("Factory.get -Scoped instance not found for ${context.scope.id} in $beanDefinition")
```
Nek-12, line 66: This is a breaking change. Make sure you increase major version before release

arnaudgiuliani: not in that case, as with such new behavior we would have ISE and can't use `getOrNull`.  Here external behavior is the same for the user, "just" a side effect internally to handle that case

Nek-12: Then this api, including exception, should be internal

arnaudgiuliani: you are right, it is still an internal API. I can tag them @KoinInternalApi in 4.1 branch.

Issue #1172, opened 2021-08-16
Using named() with enum constants crashes the app (when locale is set to Turkish)


iliaskomp:
**Describe the bug**
When users set their device language to turkish the app crashes. 

Koin is converting the enum constant to a string qualifier (lowercase string):
```
val <E : Enum<E>> Enum<E>.qualifier
    get() : Qualifier {
        return StringQualifier(this.toString().toLowerCase())
    }
```
In case that the enum constant contains the letter "I", the lower case conversion will not work since the turkish language has two "i" characters, one dotted and one dotless (see more info here https://elye-project.medium.com/remember-to-localize-for-string-conversion-726051a8958e)

**To Reproduce**
Use the following code snippet in a scratch file:
```
enum class User {
    NICK
}
val locale = Locale("tr", "TR")
User.NICK.toString().toLowerCase(locale)
```
The result of the above snippet is "nıck", where the "i" is dotless!

**Expected behavior**
The app should not crash. 
A solution would be to set a default US locale when converting to lowercase and additionally giving the option to override the default behaviour by passing the desired locale.
Perhaps there are more similar issues with other locales and the usage of an enum should be discouraged.

**Koin project used and used version (please complete the following information):**
Used in android project: io.insert-koin:koin-android:2.2.3

iliaskomp:
This issue is still valid

PR #2121, opened 2025-01-23
Enum's name as qualifier

luozejiaqun:
enum class `toString` can be override, use enum's name as its qualifier


pcherubim:
Will probably fix https://github.com/InsertKoinIO/koin/issues/1172.

PR #2165, opened 2025-04-01
Fix broken link for “Kotlin with Koin Annotations” in documentation

ykws:
It seems that the following tutorial page is no longer available:

- Kotlin with Koin Annotations

This page is not listed under the [Quick Start](https://insert-koin.io/docs/quickstart/kotlin) section anymore:

Screenshot 2025-03-05 at 16 32 30

Could you please confirm if the following page is now the correct reference?

https://insert-koin.io/docs/reference/koin-annotations/start

### Related Pull Request
- https://github.com/InsertKoinIO/koin-annotations/pull/241

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Here are the current available Koin project versions:
 
 You can find here tutorials to help you learn and get started with Koin framework:
 - [Kotlin](https://insert-koin.io/docs/quickstart/kotlin)
-- [Kotlin with Koin Annotations](https://insert-koin.io/docs/quickstart/kotlin-annotations)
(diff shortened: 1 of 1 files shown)


arnaudgiuliani:
this quickstart doesnt exists anymore ... do you think raw kotlin app demo with annotations is still required?

ykws:
@arnaudgiuliani Thanks for the feedback. I just checked and confirmed that all the tutorial links now return “Not Found.” Do you think we should remove the Tutorials section from the README? eb809751d9228c0536619416afc542d64868b24c

PR #2183, opened 2025-04-30
Continue Compose Multiplatform 1.8 version upgrade

zsmb13:
Addresses the comments in https://github.com/InsertKoinIO/koin/pull/2179#discussion_r2068665145


arnaudgiuliani:
sounds like `lifecycleViewmodelSavedState` is already covering those dependencies, right?

zsmb13:
Yes, I think that's the case

PR #2232, opened 2025-06-27
Fix scoped lookup of nested definitions (#2231)

### LINKED ISSUE #2231: Regression/4.1.0-Beta13: Scope Context changes during lookup of nested definitions
**Describe the bug**
When attempting to lookup a definition in a certain Scope which depends on another definition declared in another Scope, the resolution context changes to that other Scope rather than staying in the original one.
Consequently, nested Factory declarations cannot access elements from the original scope.

This is because the logic changed between 4.1.0 Beta 11 and 13:
from a simple [`Scope#getOrNull`](https://github.com/InsertKoinIO/koin/compare/4.1.0-Beta11...4.1.0#diff-60478c80e1f0cf8403d5a1381372f88c6d18d8e3c96411fb57f7ba13afa934c1L382) to [a more complex logic](https://github.com/InsertKoinIO/koin/compare/4.1.0-Beta11...4.1.0#diff-7323623dfbb4ec3f083c68fa2dc9d0cd0f388b3bcd2397611d2e9217eed2e51cR131-R136)
(I linked the diff to 4.1.0 as the logic changed slightly between Beta13 and the final release)

**Expected behavior**
Dependencies are looked up from other Scopes as a fallback, but the original context prevails throughout the nested lookups.

**Koin module and version:**
koin-core:4.1.0

https://github.com/SIMULATAN/koin-scope-context-reproducer

SIMULATAN:
Fixes #2231

### DIFF
--- a/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/resolution/CoreResolver.kt
+++ b/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/resolution/CoreResolver.kt
@@ -121,7 +121,18 @@ class CoreResolver(
     private fun <T> resolveFromParentScopes(scope: Scope, ctx: ResolutionContext): T? {
         if (scope.isRoot) return null
         ctx.logger.debug("|- ? ${ctx.debugTag} look in other scopes")
-        return findInOtherScope(scope,ctx)
+        return getFromLinkedScopes(scope, ctx) ?: findInOtherScope(scope, ctx)
+    }
+
+    private fun <T> getFromLinkedScopes(
+        scope: Scope,
+        ctx: ResolutionContext,
+    ): T? {
+        val parentScope = flatten(scope.linkedScopes)
+        return parentScope.firstNotNullOfOrNull {
+            ctx.logger.debug("|- ? ${ctx.debugTag} get from scope '${it.id}'")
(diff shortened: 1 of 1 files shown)


projects/core/koin-core/src/commonMain/kotlin/org/koin/core/resolution/CoreResolver.kt
```diff
@@ -131,8 +131,7 @@ class CoreResolver(
         val parentScope = flatten(scope.linkedScopes)
         return parentScope.firstNotNullOfOrNull {
             ctx.logger.debug("|- ? ${ctx.debugTag} look in scope '${it.id}'")
-            val instanceContext = if (!it.isRoot) ctx.newContextForScope(it) else ctx
-            resolveFromContextOrNull(it, instanceContext, lookupParent = false)
+            resolveFromContextOrNull(it, ctx, lookupParent = false)
```
arnaudgiuliani, line 134: Please check test impact

SIMULATAN:
Although I'm not sure if I understand the expected behavior from the test, I resolved the issue by simply adding the former logic as a fallback. Unfortunately, I cannot verify side effects locally because, for some reason, numerous tests fail..

PR #1999, opened 2024-09-24
flatten module order error

luozejiaqun:
Fixes #1998

PR #2245, opened 2025-07-12
Fix nested module overrides

domgew:
This pull request primarily adds a fix for nested module overrides.

Additionally, previously failing unit tests of the koin-core module are fixed - one even with a problem in the production code (if braces). To prevent this in the future, the projects folder's test.sh is added to the test pipeline as well.

Furthermore, the JVM version is now correctly set for the koin-core module, so that `jvmTest` now works. To reflect the necessary Java version for building for IDEs and CLIs, a corresponding `.sdkmanrc` with the same version and variant as is used in the pipeline is added.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
         uses: ./.github/actions/prepare-env
 
       - name: Install
-        run: cd projects && ./install.sh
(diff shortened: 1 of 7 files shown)


arnaudgiuliani:
Let's follow up on https://github.com/InsertKoinIO/koin/pull/1999, initially opened. For the other impact, I need to check separately

PR #2250, opened 2025-07-17
Koin + Ktor DI 3.2 Integration

lidonis:
Ktor DI Bridge Implementation
  - Added KoinDependencyMapExtension implementing Ktor 3.2's DependencyMapExtension interface
  - Registered via SPI in META-INF/services/io.ktor.server.plugins.di.DependencyMapExtension

Koin can resolve Ktor DI dependencies via KtorDIExtension resolution extension.
  - There is a potential problem with runBlocking usage

When dependency is not found, each library is delegating to the other in an infinite loop

Fixed a problem in CoreResolver that was not resolving extensions

Created a sample application to show usage.

### DIFF
--- a/examples/gradle/versions.gradle
+++ b/examples/gradle/versions.gradle
@@ -7,7 +7,7 @@ ext {
     koin_compose_version = koin_version
 
     coroutines_version = "1.9.0"
-    ktor_version = "3.1.3"//"3.2.0-eap-1310"
+    ktor_version = "3.2.1"
     jb_compose_version = "1.8.0"
 
     // Test
--- a/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/resolution/CoreResolver.kt
+++ b/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/resolution/CoreResolver.kt
@@ -75,7 +75,7 @@ class CoreResolver(
             ?: resolveFromStackedParameters(scope,instanceContext)
             ?: resolveFromScopeSource(scope,instanceContext)
             ?: resolveFromScopeArchetype(scope,instanceContext)
-            ?: if (lookupParent) resolveFromParentScopes(scope,instanceContext) else null
+            ?: (if (lookupParent) resolveFromParentScopes(scope,instanceContext) else null)
             ?: resolveInExtensions(scope,instanceContext)
     }
 
--- a/projects/ktor/koin-ktor/src/commonMain/kotlin/org/koin/ktor/di/KoinDependencyMap.kt
+++ b/projects/ktor/koin-ktor/src/commonMain/kotlin/org/koin/ktor/di/KoinDependencyMap.kt
@@ -1,26 +1,68 @@
+/*
+ * Copyright 2017-Present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.koin.ktor.di
 
-//import io.ktor.server.plugins.di.DependencyKey
-//import io.ktor.server.plugins.di.DependencyMap
-//import org.koin.core.Koin
-//import org.koin.core.qualifier.Qualifier
-//import org.koin.core.qualifier.named
-//
-///**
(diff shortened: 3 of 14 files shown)


examples/gradle/versions.gradle
```diff
@@ -7,7 +7,7 @@ ext {
     koin_compose_version = koin_version
 
     coroutines_version = "1.9.0"
-    ktor_version = "3.1.3"//"3.2.0-eap-1310"
+    ktor_version = "3.2.1"
```
arnaudgiuliani, line 10: even 3.2.2

projects/core/koin-core/src/commonMain/kotlin/org/koin/core/resolution/CoreResolver.kt
```diff
@@ -75,7 +75,7 @@ class CoreResolver(
             ?: resolveFromStackedParameters(scope,instanceContext)
             ?: resolveFromScopeSource(scope,instanceContext)
             ?: resolveFromScopeArchetype(scope,instanceContext)
-            ?: if (lookupParent) resolveFromParentScopes(scope,instanceContext) else null
+            ?: (if (lookupParent) resolveFromParentScopes(scope,instanceContext) else null)
```
arnaudgiuliani, line 78: ok

projects/ktor/koin-ktor/src/commonMain/kotlin/org/koin/ktor/di/KtorDIExtension.kt
```diff
@@ -15,25 +15,29 @@
...
+ */
+internal class KtorDIExtension(private val application : Application) : ResolutionExtension {
+    override val name: String = "ktor-di"
+    override fun resolve(scope: Scope, instanceContext: ResolutionContext): Any? {
+        val key = DependencyKey(TypeInfo(instanceContext.clazz), qualifier = instanceContext.qualifier?.value)
+        // runBlocking is required here because Ktor DI's get() function is suspend
+        // The blocking call is generally safe as dependency resolution is typically fast and non-blocking
+        // WARNING: This may cause problems for users as it can impact performance
+        return runBlocking {
+            application.dependencies.get(key)
```
arnaudgiuliani, line 40: @lidonis we can look at getDeferred or something

projects/ktor/koin-ktor/src/commonMain/kotlin/org/koin/ktor/di/KoinDependencyMap.kt
```diff
@@ -1,26 +1,68 @@
...
+}
+
+/**
+ * Koin implementation of DependencyMap interface
+ */
+class KoinDependencyMap(private val koin: Koin) : DependencyMap {
+
+    override fun contains(key: DependencyKey): Boolean =
+        try {
+            resolve(key)
```
arnaudgiuliani, line 47: Let me check if we have a lookup check here, else we are directly resolving the key

projects/ktor/koin-ktor/src/commonMain/kotlin/org/koin/ktor/di/KtorDIExtension.kt
arnaudgiuliani, line 40: @osipxd we can't use getDeferred here, as we don't have reified type info from this point. we only have KClass.
Not sure runBlocking is the good way. Else we can go with typeOf() or something ?

bjhham: The key construction using the class reference looks right since the reified type's not available.

You can use `application.dependencies.getBlocking(key)` here to let Ktor handle the blocking part.

In the case of property delegates, we also have a step to indicate that the key should be checked during startup validation, which might be applicable elsewhere:

```kotlin
public inline operator fun <reified T> provideDelegate(
    thisRef: Any?,
    prop: KProperty<*>
): ReadOnlyProperty<Any?, T> {
    val key = DependencyKey<T>()
        .also(::require)
    return ReadOnlyProperty { _, _ ->
        getBlocking(key)
    }
}
```

bjhham (commented):
👍

projects/ktor/koin-ktor/src/commonMain/kotlin/org/koin/ktor/di/KtorDIExtension.kt
arnaudgiuliani: Ok, interesting. Thanks

arnaudgiuliani: @bjhham one question: if Ktor DI is falling back on Koin and Koin falling back on Ktor DI, we might go with looping when none of the DIs have a given definition

bjhham:
Hey @lidonis you can use `getBlocking` on other platforms, but it can only be used with deferred processing, so lazy retrieval and access after the server is started.  It is this way due to the lack of `runBlocking` available in JS/WASM targets.

PR #2255, opened 2025-07-21
refactor(androidx): standardize Koin startup package naming

MycroftWong:
Rename `org.koin.androix.startup` to `org.koin.androidx.startup` (correcting“androix” to official “androidx” naming)


kibettheophilus (approved):

arnaudgiuliani:
can you check breaking impact? I was looking at it, but seems it may breaks people's app :/

Issue #2292, opened 2025-10-19
sharedKoinViewModel fails with type-safe navigation routes


nazaburo:
**Describe the bug**
The sharedKoinViewModel function in koin-compose-viewmodel-navigation doesn't work correctly with [type-safe navigation](https://developer.android.com/guide/navigation/design/type-safety) (serializable routes). When the default parameter is used, it causes a runtime crash because the serializable route object from destination.parent?.route is incorrectly treated as a String route by NavController.getBackStackEntry(), leading to a type mismatch.

**Environment**
  - Koin version: 4.1.1
  - Navigation Compose version: 2.9.5
  - Kotlin version: 2.2.20
  - Compose BOM: 2025.10.00

**The Problem**
When using type-safe navigation with @Serializable routes and the default parameter navGraphRoute: Any? = this.destination.parent?.route, the function receives a serializable route object but NavController.getBackStackEntry() internally treats it as if it were a String route, causing a type mismatch.

The navigation graph was created with serializable route objects, but getBackStackEntry() tries to match it as a String representation, so the entry cannot be found and crashes.

**Sample Project**
I've created a minimal reproduction project that demonstrates this issue:
https://github.com/nazaburo/SharedKoinViewModelSample

**Proposed Fix**
I've submitted a pull request with a suggested fix:
https://github.com/nazaburo/SharedKoinViewModelSample/pull/2

The fix properly handles both String-based and serializable routes by detecting the route type and calling the appropriate getBackStackEntry() overload.

aurimas-zarskis:
Hey, I ran into same issue, but in my case it crashes with both type-safe routes and string based routes.
This is the navigation setup I was testing with:
```
navigation(route = "root", startDestination = "screen1") {
    composable(route = "screen1") {
        val viewModel = it.sharedKoinViewModel<SharedViewModelTest>(mainNavController)
        Screen(viewModel) {
            mainNavController.navigate("screen2")
        }
    }
    composable(route = "screen2") {
        val viewModel = it.sharedKoinViewModel<SharedViewModelTest>(mainNavController)
        Screen2(viewModel)
    }
}
````
Then I used this to navigate: `mainNavController.navigate("root")`
After navigation the app crashed with this error: `Destination with route String cannot be found in navigation graph ComposeNavGraph`
I'm using compose multiplatform btw

arnaudgiuliani:
let's see in your PR

PR #2293, opened 2025-10-19
Fix sharedKoinViewModel fails with type-safe navigation routes

nazaburo:
Fix sharedKoinViewModel fails with [type-safe navigation](https://developer.android.com/guide/navigation/design/type-safety) routes #2292

### DIFF
--- a/examples/build.gradle
+++ b/examples/build.gradle
@@ -25,6 +25,7 @@ buildscript {
 plugins {
     // Apply the new Kotlin Compose plugin
     id 'org.jetbrains.kotlin.plugin.compose' version "$kotlin_version" apply false
+    id 'org.jetbrains.kotlin.plugin.serialization' version "$kotlin_version" apply false
 }
 
 wrapper {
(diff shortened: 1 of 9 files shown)


arnaudgiuliani:
I don't see why it bring stronger verification here, as the compiler should already help here.
I believe more in a K 2.2 compat problem.

nazaburo:
@arnaudgiuliani 
I've added test cases to `sample-android-compose` demonstrating the fix.

**Added:**
- Type-Safe navigation routes (`NavigationGraphRoute`, `FirstScreenRoute`, `SecondScreenRoute`)
- Two navigation examples:
  - **Type-Safe Navigation** (`navigationTypeSafeGraph`) - Uses `@Serializable` route objects
  - **String-based Navigation** (`navigationStringGraph`) - Uses traditional string routes
- Two screens sharing a `SharedViewModel` via `sharedKoinViewModel`
- Simple ViewModel to verify the same instance is shared between screens

You can test it by running the app and tapping "(>)Type Safe Navigation Example" or "(>)String Navigation Example". Both examples demonstrate that screens share the same ViewModel
instance (shown by matching instance IDs and shared timestamp), confirming the fix works correctly with Type-Safe routes.

arnaudgiuliani:
Thanks for your work @nazaburo 🙏

PR #2339, opened 2026-01-07
Update limitations section in navigation3.md

BierDav:
Added note about the inability to set fallback in Navigation 3 limitations.


arnaudgiuliani:
what do you expect from fallback? is it limitations of Nav3 or Koin?

BierDav:
The backstack could contain a value that is not declared in the nav routes definition, fallback will be rendered in this case

It is a limitation of Koin, because currently it is not supported to define a fallback

arnaudgiuliani:
yes to be reviewed, where we can more deeper to help on it

arnaudgiuliani:
I can take a look at the API to help fix it

PR #2375, opened 2026-03-08
Use `@ReadOnlyComposable` wherever possible

Miha-x64:
P. S. `org.koin.compose.viewmodel.{koinViewModel, koinActivityViewModel}` also wanna be `@ReadOnlyComposable` but `androidx.lifecycle.viewmodel.compose.LocalViewModelStoreOwner.current` doesn't want them to be happy.


Miha-x64:
NB: as it turns out, the proposed `@ReadOnlyComposable` now has no actual effect https://issuetracker.google.com/issues/335877194

PR #2389, opened 2026-03-29
fix: add volatile and test for concurrency

inemtsev:
After investigating some concurrency issues we've had with version 4.1.1, I realized most of these were fixed with 4.2.0.

But, I believe this one can still cause an issue in rare cases.  **_closed** should be volatile because checkScopeIsOpen() reads _closed outside any lock the JVM is free to let other threads see a stale cached value of false even after close() has fully returned.

It was hard to create a test, but I managed to reproduce with some AI help.

Adding volatile fixes the failing test


inemtsev:
@arnaudgiuliani Do you think we can include this in the next patch version?

arnaudgiuliani:
will be in 4.2.1 👍

PR #2399, opened 2026-04-08
fix(viewmodel): keep useIndexedValues in AndroidParametersHolder

0xZhangKe:
## Summary

Fix ViewModel parameter wrapping to preserve `ParametersHolder.useIndexedValues`.

## Details

`parameterArrayOf(...)` sets `useIndexedValues = true`, but this flag was lost when parameters were wrapped by `AndroidParametersHolder` during ViewModel resolution.

This change keeps the original flag so indexed parameter semantics are preserved.


0xZhangKe:
Hi @arnaudgiuliani , gentle ping on this one.

This fixes parameterArrayOf(...) semantics being lost when ViewModel parameters are wrapped by AndroidParametersHolder. 

The PR is small and currently mergeable.
Would you mind taking a look when you get a chance? Thanks!

PR #2423, opened 2026-05-25
Add koinGet() composable and deprecate koinInject() for naming consistency

### LINKED ISSUE #2405: org.koin.compose.koinInject has inconsistent naming with Scope functions
Despite `inject` in the name `koinInject` uses `Scope.get()` under the hood.
Having `inject` in the name suggests that the dependency is lazily resolved.

Current `koinInject` should rather be called `koinGet`.

Would be nice to have a parallel `koinInject` implementation which would use `Scope.inject()` under the hood and return `Lazy<T>` instead of `T`.

Also, as the Composables use `remember` under the hood they could be also named like `rememberKoinGet()` and `rememberKoinInject()`, so it's more obvious to the user.

saadkhalidkhan:
## Summary

- Add `koinGet()` composable functions as the correctly-named alternative to `koinInject()`
- Deprecate `koinInject()` with `@Deprecated` + `ReplaceWith` for easy IDE migration
- Update Compose docs to use `koinGet()`

### Rationale

In Koin core:
- `KoinComponent.get()` / `Scope.get()` = eager resolution, returns `T`
- `KoinComponent.inject()` / `Scope.inject()` = lazy resolution, returns `Lazy<T>`

In Compose, `koinInject()` calls `Scope.get()` (eager) but the name suggests lazy resolution.
`koinGet()` aligns the Compose API naming with core Koin conventions.

Fixes #2405

## Test plan

- [x] `koinGet()` functions mirror all three overloads of `koinInject()`
- [x] Deprecated `koinInject()` delegates to `koinGet()` (no behavior change)
- [x] Built `:compose:koin-compose:compileDebugKotlinAndroid` successfully
- [x] Updated compose docs to reference `koinGet()`

### DIFF
--- a/docs/reference/koin-compose/compose-scopes.md
+++ b/docs/reference/koin-compose/compose-scopes.md
@@ -29,7 +29,7 @@ fun FeatureScreen() {
 @Composable
 fun FeatureContent() {
     // Resolves from parent KoinScope
-    val cache = koinInject<FeatureCache>()
+    val cache = koinGet<FeatureCache>()
 }
 ```
 
@@ -61,7 +61,7 @@ NavHost(navController, startDestination = "home") {
 @Composable
 fun DetailScreen() {
     // Dependencies scoped to this navigation destination
-    val repository = koinInject<ScreenRepository>()
+    val repository = koinGet<ScreenRepository>()
     val viewModel = koinViewModel<ScreenViewModel>()
 }
 ```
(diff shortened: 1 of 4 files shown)


Nailik:
will `koinGet` also support `@Provided`?  Currently that is an issue in typesafety check `KOIN-D002 ` is thrown if a different modul provides the implementation and the modul that calls `koinInject` only knows the interface.

PR #2432, opened 2026-06-03
Use a monotonic counter for Ktor request scope ids

### LINKED ISSUE #2410: `RequestScope` UUID blocks Netty event loop on every request
> Note: this isn't a KFIP-scale design, it's a small behavioral change to existing per-request scope wiring.

The koin-ktor `Koin` plugin installs a per-request scope hook in `setupKoinScope`:

```kotlin
// org/koin/ktor/plugin/KoinPlugin.kt (4.2.1)
on(CallSetup) { call ->
}
```

`RequestScope.<init>` generates a UUID via `KoinPlatformTools.generateId()`:

```kotlin
// org/koin/ktor/plugin/RequestScope.kt (4.2.1)
class RequestScope(private val _koin: Koin, call: ApplicationCall) : KoinScopeComponent {
    private val scopeId = "request_" + KoinPlatformTools.generateId()
    override val scope = createScope(scopeId = scopeId, source = call)
}
```

On JVM, `generateId()` resolves to `Uuid.random().toString()`, which calls `SecureRandom.nextBytes(...)`. With the default JCA provider on Linux (`NativePRNG`), this reads from `/dev/urandom` via `FileInputStream.readBytes`.

When the Ktor engine is Netty, the read lands on a Netty event-loop thread, which BlockHound surfaces as an event-loop blocking violation.

This happens **on every HTTP request**, even when the application declares no `scoped { }` definitions, `setupKoinScope` is installed unconditionally.

**Reproduction (JDK 21, Linux/macOS):**

- Ktor 3.4.0 + `ktor-server-netty`
- `koin-ktor` 4.2.1
- `BlockHound.install()` before the engine starts
- Send GET requests to any route

Stack trace:
```
reactor.blockhound.BlockingOperationError: Blocking call! java.io.FileInputStream#readBytes
    at java.base/sun.security.provider.NativePRNG$RandomIO.readFully
    at java.base/java.security.SecureRandom.nextBytes
    at kotlin.uuid.UuidKt__UuidJVMKt.secureRandomBytes(UuidJVM.kt:20)
    at kotlin.uuid.Uuid$Companion.random(Uuid.kt:582)
    at org.koin.mp.KoinPlatformTools_jvmKt.generateId(KoinPlatformTools.jvm.kt:7)
    at org.koin.ktor.plugin.RequestScope.<init>(RequestScope.kt:31)
    at org.koin.ktor.plugin.KoinPluginKt$setupKoinScope$1.invokeSuspend(KoinPlugin.kt:97)
    ...
    at io.ktor.server.netty.NettyApplicationCallHandler.handleRequest(NettyApplicationCallHandler.kt:55)
    at io.netty.channel.SingleThreadIoEventLoop.run
```

Avoid the SecureRandom-backed UUID per request when it isn't needed. Two complementary directions worth considering:

- Make per-request scope creation **opt-in or lazy**, so applications that don't use `scoped { }` definitions don't pay for the wiring at all.
- Reconsider whether `RequestScope` IDs need to be **cryptographically random**. They only need uniqueness within a process, so a non-blocking source would be enough.

- Dropping the `koin-ktor` plugin and replacing it with a small `KoinPlatform.getKoin()`-backed shim that mirrors `Route.inject<T>()` / `Application.getKoin()`. Works when no `scoped { }` definitions are used, but discards the per-request scope feature entirely, not a viable option for users who rely on it.

**Target Koin project**

`koin-ktor` (verified on 4.2.1).

lfavreli-betclic:
## Problem

Fixes #2410.

`RequestScope` builds its scope id with `KoinPlatformTools.generateId()`, a SecureRandom-backed UUID. That blocking call runs on every request, which is problematic on non-blocking / event-loop threads. It happens even when no `scoped { }` definitions are declared, since the request scope is created unconditionally.

## Change

Replace the random UUID with a process-wide monotonic `AtomicLong`, seeded with the current wall-clock millis so ids don't restart at the same value across process restarts:

`request_<n>` (e.g. `request_1780500226902`)

A request scope id only needs uniqueness within the running process, which the counter guarantees without touching `SecureRandom`.

- `kotlin.concurrent.atomics.AtomicLong` (multiplatform, `RequestScope` is `commonMain`)
- `kotlin.time.Clock` for the seed (stable since Kotlin 2.3, no new dependency)

_Reference_: same approach backs block id generation in [Apache Spark](https://github.com/apache/spark/blob/13b526d534e3b6115be4769d39d861b321bb3a02/streaming/src/main/scala/org/apache/spark/streaming/receiver/ReceiverSupervisorImpl.scala#L95).

## Test

Added to `KoinPluginRunTest`: 100 requests against the running plugin assert every request scope id is unique and counter-formatted.

### DIFF
--- a/projects/ktor/koin-ktor/src/commonMain/kotlin/org/koin/ktor/plugin/RequestScope.kt
+++ b/projects/ktor/koin-ktor/src/commonMain/kotlin/org/koin/ktor/plugin/RequestScope.kt
@@ -19,16 +19,25 @@ import io.ktor.server.application.ApplicationCall
 import org.koin.core.Koin
 import org.koin.core.component.KoinScopeComponent
 import org.koin.core.component.createScope
-import org.koin.mp.KoinPlatformTools
-import org.koin.mp.generateId
+import kotlin.concurrent.atomics.AtomicLong
+import kotlin.concurrent.atomics.ExperimentalAtomicApi
+import kotlin.concurrent.atomics.incrementAndFetch
+import kotlin.time.Clock
 
 /**
  * Request Scope Holder
  *
  * @author Arnaud Giuliani
+ * @author Loïc Favreliere
  */
+@OptIn(ExperimentalAtomicApi::class)
 class RequestScope(private val _koin: Koin, call: ApplicationCall) : KoinScopeComponent {
-    private val scopeId = "request_"+KoinPlatformTools.generateId()
+    private val scopeId = "request_" + counter.incrementAndFetch()
     override fun getKoin(): Koin = _koin
     override val scope = createScope(scopeId = scopeId, source = call)
+
+    private companion object {
+        // Monotonic counter seeded with the current time, for process-unique request scope ids
+        private val counter = AtomicLong(Clock.System.now().toEpochMilliseconds())
+    }
 }
--- a/projects/ktor/koin-ktor/src/jvmTest/kotlin/org/koin/ktor/ext/KoinPluginRunTest.kt
+++ b/projects/ktor/koin-ktor/src/jvmTest/kotlin/org/koin/ktor/ext/KoinPluginRunTest.kt
@@ -27,6 +27,10 @@ import io.ktor.server.response.respond
 import io.ktor.server.routing.get
 import io.ktor.server.routing.routing
 import io.ktor.server.testing.testApplication
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.runBlocking
 import org.junit.Before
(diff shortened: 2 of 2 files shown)


inemtsev:
Look good so far, I double checked and AtomicLong is actually more safe than UUID for this case, since UUID is probabilistic. But tests are not sufficient, a sequential loop can't produce a data race, so the test passes regardless of thread-safety. So the PR's test would still go green even if you had used a thread-unsafe counter. 

Can you add a more concurrent test?
Something like this?

```kotlin
val ids = coroutineScope {
          (1..1_000)
              .map { async(Dispatchers.Default) { client.get("testurl").headers["X-Scope-Id"] } }
              .awaitAll()
      }
```

lfavreli-betclic:
Good catch, @inemtsev. Added a concurrent pass (1000 requests via `async(Dispatchers.Default)`) alongside the sequential one and assert global uniqueness.

How about it?

inemtsev:
Look good with me. 

@arnaudgiuliani Can you please take a look?

arnaudgiuliani:
@claude review it

arnaudgiuliani:
> Look good with me.
> 
> @arnaudgiuliani Can you please take a look?

Thanks, will take a look and propose for 4.3

arnaudgiuliani:
Thanks a lot for this @lfavreli-betclic 🙏

Since #2410 is milestoned for **4.2.2**, I've cherry-picked your two commits onto the `4.2.2` release branch — **with your authorship preserved** — in #2445 (verified: ktor jvmTest green, commonMain compiles on iOS + wasmJs, apiCheck clean).

arnaudgiuliani:
Reopening to re-run CI against the new `4.2.2` base — the previous red checks were from the original run against `main` (2026-06-04) and weren't refreshed when the base changed. Verified locally on 4.2.2: ktor jvmTest green, commonMain compiles on iOS + wasmJs, apiCheck clean.

Issue #2387, opened 2026-03-25
Can't get SavedStateHandle not in constructor of ViewModel in 4.2.0


Grouen:
**Describe the bug**
In Koin 4.1.1 it was possible to get SavedStateHandle not only in constructor of the ViewModel.
Like this
```
class MyViewModel : ScopeViewModel() {
    private val savedStateHandle: SavedStateHandle = get()
}
```
Now it throws `No definition found for type 'androidx.lifecycle.SavedStateHandle'`.
I checked the code, and it looks like `CoreResolverV2.resolveFromLinkedScopes` doesn't allow to reach `AndroidParametersHolder.getOrNull`. (Actual logic of dropping scope happens in `findDefinitionInScope`)

**To Reproduce**
Get this kind of view model in the any app component.
```
class MyViewModel : ScopeViewModel() {
    private val savedStateHandle: SavedStateHandle = get()
}
```

**Expected behavior**
Koin should get the available SavedStateHandle from the linked scope as it was before

**Koin module and version:**
 `koin-core:4.2.0`

arnaudgiuliani:
Hi, thanks for the report. We investigated this and have some findings:

**`scope.get<SavedStateHandle>()` in a ViewModel property initializer is not a supported pattern.**

`SavedStateHandle` is provided by `AndroidParametersHolder`, which intercepts `getOrNull(SavedStateHandle::class)` and creates it from `CreationExtras`. These parameters are stacked on the **factory scope** (the scope used by `KoinViewModelFactory`), not on the ViewModel's own scope created by `ScopeViewModel.viewModelScope()`.

When you do `scope.get<SavedStateHandle>()` in a `ScopeViewModel` property, the `get()` runs on the ViewModel's own scope — which has no access to the factory's stacked parameters. We verified this is the case in both 4.1.1 and 4.2.0.

**The correct approach is constructor injection:**
```kotlin
class MyViewModel(savedStateHandle: SavedStateHandle) : ViewModel()
```

This way `SavedStateHandle` is resolved via `AndroidParametersHolder` during the ViewModel factory's `getWithParameters` call, where the parameters are available.

Could you double-check what your code looked like in 4.1.1? Is it possible you were using constructor injection before and switched to `scope.get()` around the same time as the upgrade?

Also note: `ScopeViewModel` + `viewModelScopeFactory()` option should not be combined — it's one or the other, as both create their own scope.

Grouen:
[...]

Issue #2370, opened 2026-03-03
Parameter overrides a qualified dependency


mtyy:
**Describe the bug**
Instead of a qualified dependency from root scope, Koin injects a parameter.

**To Reproduce**
1) Have additional scope
2) The type of the conflicting dependencies must be the same (String in this case).

**Expected behavior**
It should find the qualified dependency, as specified in the factory. It should not silently use whatever it finds of the same type (**especially** as its not qualified).

**Koin module and version:**
Happens on 4.1.1
Did not happen on 4.0.0 nor 3.5.6

**Snippet or Sample project to help reproduce**
```kotlin
import org.junit.Assert.assertEquals
import org.junit.Test
import org.koin.core.context.startKoin
import org.koin.core.parameter.parametersOf
import org.koin.core.qualifier.named
import org.koin.dsl.module
import org.koin.test.KoinTest

private const val GLOBAL_CONFIG_STRING = "CONFIG_STRING"
private val ConfigQualifier = named("Config")
private val ActivityScope = named("TestScope")

class Service(val config: String)

data class UseCase(val parameter: String, val service: Service)

class QualifiedShadowingBugTest : KoinTest {

    private val testModule = module {
        single(ConfigQualifier) { GLOBAL_CONFIG_STRING }

        factory { Service(config = get(ConfigQualifier)) }

        factory { (parameterString: String) ->
            UseCase(parameter = parameterString, service = get())
        }

        scope(ActivityScope) { }
    }

    @Test
    fun `qualified String should NOT be shadowed by parametersOf value`() {
        startKoin { modules(testModule) }
        val scope = getKoin().createScope("test", ActivityScope)

        val useCase: UseCase = scope.get { parametersOf("parameterString") }

        assertEquals("parameterString", useCase.parameter)
        // failure: useCase.service.config value is also "parameterString"
        assertEquals(GLOBAL_CONFIG_STRING, useCase.service.config)
    }

}
```

arnaudgiuliani:
Fixed in #2437, landing in **4.2.2**.

A qualified `get(named(...))` no longer reads injected/stacked parameters — parameters carry no qualifier, so a value passed via `parametersOf` of the same type can no longer shadow a qualified dependency (resolves from the graph as in 4.0.0/3.5.6). Covered by a regression test built from your repro 🙏

Closing — please test against the next **4.2.2** build/snapshot once #2437 is merged, and reopen if it persists.

PR #2440, opened 2026-06-11
Fix #2406, #2435: qualified requests no longer pick up stacked parameters

### LINKED ISSUE #2406: Broken checkKoinModules functionality on get<String>
**Describe the bug**
InsertKoinIO/koin#2330 has changed the order singletons are introduced to the scope and now the
`MockParameter` is taken into scope before actual configuration. So any named <String> singletons are replaced by empty `""`

**To Reproduce**
Steps to reproduce the behavior (this worked ok in 4.1.1, and is now broken in 4.2.1) are in the snippet below.
Basically the named string singletons are ignored and replaced by empty strings, causing the failures in other beans. 

**Expected behavior**
A way to properly check (in unit tests) the wiring in the application. And `verify()` does not cover the needs of such check. 

**Koin module and version:**
`koin-core:4.2.1`

```kotlin
class MyService(env: String) {
    init { require(env.isNotEmpty()) { "env was empty!" } }
}

fun appModules() = listOf(module {
    single(named("env")) { System.getenv("MY_ENV") ?: error("MY_ENV not defined") }
    single { MyService(get(named("env"))) }
})

class AppModulesTest {
    @Test
    fun checkWiring() {
        checkKoinModules(appModules() + module {
            single(named("env")) { "TEST" }
        })
    }
}

```

### LINKED ISSUE #2435: CoreResolverV2 using injected param instead of named
**Describe the bug**
When having the same type for `InjectedParam` and `Named`, koin initializes the class with the injected param.

**To Reproduce**
```kotlin
@Factory
class Class1(
    @InjectedParam private val id: String,
    private val class2: Class2,
)

@Factory
internal class Class2(
    @Named("customId") private val id: String,
) 

private val class1: Class1 by inject {
        parametersOf("event-id")
    }
```

This will initialize both `Class1` and `Class2` with "event-id". With koin 4.1.1 and `CoreResolver` it will initialize `Class2` with the named id.

**Expected behavior**
Named parameters should not be initialized with value from `ParametersHolder`.

**Koin module and version:**
 koin-android: 4.1.1 -> 4.2.1
koin-annotation, koin-ksp-compiler: 2.3.1

ChristophyBarth:
## Problem
A `get(named(...))` could be satisfied by a same-typed value sitting on the scope's
parameter stack, shadowing the qualified definition. So passing `parametersOf("event-id")`
to one dependency could bleed that String into an unrelated `@Named` String further down
the graph (#2435), and `checkKoinModules` handed `get(named("env"))` the MockParameter
default `""` instead of the named single (#2406).

## Cause
In `CoreResolverV2`, resolution tries the parameter stack before the registry. Parameters
match by type only and carry no qualifier, so a qualified request would match whatever
same-typed value was on the stack.

## Fix
`resolveFromStackedParameters` now returns null when the request has a qualifier, letting
it fall through to the registry. This mirrors the rule already used by
`resolveFromScopeSource`.

## Behavioral change
Qualified resolutions could previously return a stacked parameter of the matching type;
they now always resolve from definitions. Unqualified resolution, including the #2387
stacked-parameter path, is unchanged.

## Tests
Added cases in `ParametersInjectionTest` (root, scope, and linked-scope variants) and
`CheckModulesTest`. Verified on JVM, wasmJs, and native.

Closes #2406
Closes #2435

### DIFF
--- a/docs/reference/koin-core/injection-parameters.md
+++ b/docs/reference/koin-core/injection-parameters.md
@@ -80,6 +80,9 @@ val myModule = module {
 }
 ```
 
+:::note
+Graph resolution of injected parameters only applies to unqualified requests: a qualified request like `get(named("a_name"))` always resolves from your module definitions, never from injected parameters — parameters are matched by type and carry no qualifier.
+:::
 
 ## Injected parameters: indexed values or set (`3.4.3`)
 
--- a/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/resolution/CoreResolverV2.kt
+++ b/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/resolution/CoreResolverV2.kt
@@ -27,6 +27,9 @@ import org.koin.ext.getFullName
 
 /**
  * Resolver - optimised version
+ *
+ * Resolution order: injected parameters, stacked parameters (unqualified requests only),
+ * registry (direct definition, scope archetype, scope source, linked scopes), extensions.
  */
 @KoinInternalApi
 class CoreResolverV2(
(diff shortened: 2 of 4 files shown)


ChristophyBarth:
Closing — this was fixed independently on main by 4fae618a (#2370/#2408), which makes the identical change in CoreResolverV2. I'll open a small follow-up PR with the docs note and the checkmodules-path test coverage that aren't on main yet.

Issue #2417, opened 2026-05-15
Compose Multiplatform injection of ViewModel with SavedStateHandle fails with 4.2.0-beta3 onwards


realdadfish:
In 4.2.0-beta3 the resolution of ViewModels such as 

```kotlin
class MyViewModel(val savedStateHandle: SavedStateHandle): ViewModel() { ... }
```

changed. Previously

```kotlin
module {
   viewModelOf(::MyViewModel)
}
```

accompanied with

```kotlin
@Composable
fun MyComposable(vm: MyViewModel = koinViewModel())
```

was enough to create and instantiate `MyViewModel` correctly. With 4.2.0-beta3 a new requirement popped in to wrap `viewModelOf` into a `viewModelScope {}` to still have `SavedStateHandle` properly resolved, otherwise this popped up:

```
java.lang.IllegalArgumentException: CreationExtras must have a value by `SAVED_STATE_REGISTRY_OWNER_KEY`
	at androidx.lifecycle.SavedStateHandleSupport.createSavedStateHandle(SavedStateHandleSupport.kt:95)
	at org.koin.viewmodel.factory.AndroidParametersHolder.getOrNull(AndroidParametersHolder.kt:48)
	at org.koin.core.resolution.CoreResolverV2.resolveFromContextOrNull(CoreResolverV2.kt:197)
	at org.koin.core.resolution.CoreResolverV2.resolveFromContext(CoreResolverV2.kt:43)
	at org.koin.core.scope.Scope.resolveFromContext(Scope.kt:330)
	at org.koin.core.scope.Scope.stackParametersCall(Scope.kt:293)
	at org.koin.core.scope.Scope.resolveInstance(Scope.kt:279)
	at org.koin.core.scope.Scope.resolve(Scope.kt:252)
	at org.koin.core.scope.Scope.get(Scope.kt:234)
	at ...
```

So I did as I was told and wrapped my `viewModelOf` inside the proper scope:

```kotlin
module {
    viewModelScope {
        viewModelOf(::MyViewModel)
    }
}
```

This now (tested with 4.2.0-beta3 and 4.2.1) however leads to the following test and production runtime exception:

```
No definition found for type 'org.example.MyViewModel' on scope '['_root_']'.. Check or add definition for type 'org.example.MyViewModel' in scope '_root_'.
        at org.koin.core.resolution.CoreResolverV2.resolveFromContext(CoreResolverV2.kt:185)
[...]

Issue #2393, opened 2026-03-31
Inconsistent behavior of resolver with 4.1.1 (CoreResolver) v.s. 4.2 (CoreResolverV2)


b95505017:
Not sure which one is WAI 🤔 

**CoreResolver**
```kotlin
    private fun <T> resolveFromContextOrNull(scope : Scope, instanceContext: ResolutionContext, lookupParent : Boolean = true): T? {
        return resolveFromInjectedParameters(instanceContext)
            ?: resolveFromRegistry(scope,instanceContext)
            ?: resolveFromStackedParameters(scope,instanceContext)
            ?: resolveFromScopeSource(scope,instanceContext)
            ?: resolveFromScopeArchetype(scope,instanceContext)
            ?: (if (lookupParent) resolveFromParentScopes(scope,instanceContext) else null)
            ?: resolveInExtensions(scope,instanceContext)
    }
```

**CoreResolverV2**
```kotlin
    private fun <T> resolveFromContextOrNull(scope : Scope, instanceContext: ResolutionContext): T? {
        return resolveFromInjectedParameters(instanceContext)
            ?: resolveFromStackedParameters(scope, instanceContext)
            ?: resolveFromRegistry(scope, instanceContext)
            ?: resolveInExtensions(scope, instanceContext)
    }
```

The biggest difference between v1 & v2 is the order of `resolveFromStackedParameters` v.s. `resolveFromRegistry`.
If I declare a singleton with `MutableStateFlow(A)`, then `get` from somewhere in companion with `parametersOf(MutableStateFlow(B))`, I could extract correctly with `p[0]` & `get` respectively in 4.1.1, but in 4.2 I could only get `MutableStateFlow(B)` in either way.

arnaudgiuliani:
Resolved across the 4.2.2 work:

- The **qualified** half — a `parametersOf` value shadowing a qualified `get(named(...))` of the same type — was a real bug and is **fixed** (#2370 / #2408, merged for 4.2.2): qualified lookups no longer read stacked/injected parameters.
- The **unqualified** half — passing `parametersOf(MutableStateFlow(B))` returns that value for an unqualified `get<MutableStateFlow<…>>()` even when a singleton of the same type exists — is **by design**: a `parametersOf` value of the same type as the request is returned directly (the V1→V2 ordering change). Now documented in the [injection parameters reference](https://insert-koin.io/docs/reference/koin-core/injection-parameters). Use a **qualifier** or a **wrapper type** to disambiguate.

[...]

Issue #2312, opened 2025-11-20
[Kotlin 2.3.0] Js/WasmJs PlatformTools to use `qualifiedName` (instead of `simpleName`)


zombiehub:
**Describe the bug**
Please switch to `qualifiedName` on Js/WasmJs -- since using simple class name clobbers bean definitions (and trips up `allowOverrides = false`), causing catastrophic failures in Browser vs Mobile/Desktop.

It looks like Kotlin 2.3 will have `-Xwasm-kclass-fqn` by default. See https://youtrack.jetbrains.com/issue/KT-69621/K-Wasm-Consider-enabling-support-for-KClass.qualifiedName-by-default

Thanks.

arnaudgiuliani:
Yes, 2.3 will be part of Koin 4.2

arnaudgiuliani:
Fixed in 4.2 beta3

zombiehub:
Sorry, spoke too soon. Seems like the problem persists... I get:

```
(Koin) (+) override index 'InAuthScope::_auth_' -> '[Scoped: 'InAuthScope',scope:_auth_]'
```

indicating that short names are still being used. And of course, the web crashes afterwards because the type will be wrong.

arnaudgiuliani:
let's check that 👍

arnaudgiuliani:
Status: the fix for **WasmJs** is a one-liner — `KoinPlatformTools.getClassName` should prefer `qualifiedName` (now available on WasmJs since Kotlin 2.3.0) over `simpleName`, so same-simple-name classes stop colliding on the type index key. Verified locally: with two same-simple-name classes a request returned the wrong type (`ClassCastException`) before, and resolves correctly after.

**Holding it until it can ship with a regression test**, which is currently blocked:
- **Kotlin/JS can't be fixed the same way** — `KClass.qualifiedName` is a *compile error* on JS ("This reflection API is not supported in Kotlin/JS"), so JS keeps `simpleName`. A `commonTest` test therefore fails on the JS target.
- A **WasmJs-only test source set crashes the Kotlin/Wasm compiler** (`ArrayIndexOutOfBoundsException` in `compileTestKotlinWasmJs`), and koin-core has no per-target test source sets.

Moving to **4.3.0**: ship the WasmJs fix once there's a viable regression-test path (resolve the wasmJsTest compiler crash, or a Kotlin version that fixes it). JS remains limited by upstream (no `qualifiedName`).

Issue #2044, opened 2024-11-01
No definition found for type 'androidx.lifecycle.SavedStateHandle'


juan-wesper:
**Describe the bug**
This used to work fine for v3.5.6, but we've seen this after upgrading to v4.0.0.

**To Reproduce**
1. Have a ViewModel that takes `SavedStateHandle` as param.
2. Use `viewModelOf` (the one from the new package) on your module.
3. Turn 'Don't keep activities' from the Developer Options on your phone settings.
4. Open the screen in which the aforementioned ViewModel is used.

Note that this does not always happen

**Expected behavior**
The ViewModel is instantiated correctly

**Actual behavior**
The app crashes with the following exception:
```
org.koin.core.error.InstanceCreationException: Could not create instance for '[Factory: '<redacted>.ui.screens.test.upload.UploadViewModel']'
at org.koin.core.instance.InstanceFactory.create(InstanceFactory.kt:58)
at org.koin.core.instance.FactoryInstanceFactory.get(FactoryInstanceFactory.kt:38)
at org.koin.core.registry.InstanceRegistry.resolveInstance$koin_core(InstanceRegistry.kt:110)
at org.koin.core.scope.Scope.resolveFromRegistry(Scope.kt:321)
at org.koin.core.scope.Scope.resolveFromContext(Scope.kt:311)
at org.koin.core.scope.Scope.stackParametersCall(Scope.kt:281)
at org.koin.core.scope.Scope.resolveInstance(Scope.kt:259)
at org.koin.core.scope.Scope.resolveWithOptionalLogging(Scope.kt:232)
at org.koin.core.scope.Scope.get(Scope.kt:215)
at org.koin.viewmodel.factory.KoinViewModelFactory.create(KoinViewModelFactory.kt:39)
at androidx.lifecycle.viewmodel.ViewModelProviderImpl_androidKt.createViewModel(ViewModelProviderImpl.android.kt:34)
at androidx.lifecycle.viewmodel.ViewModelProviderImpl.getViewModel$lifecycle_viewmodel_release(ViewModelProviderImpl.kt:65)
at androidx.lifecycle.viewmodel.ViewModelProviderImpl.getViewModel$lifecycle_viewmodel_release$default(ViewModelProviderImpl.kt:47)
at androidx.lifecycle.ViewModelProvider.get(ViewModelProvider.android.kt:91)
at org.koin.viewmodel.GetViewModelKt.resolveViewModel(GetViewModel.kt:60)
[...]

PR #2441, opened 2026-06-11
Improve SavedStateHandle / ViewModel DX (#2044, #2417) — actionable errors + R8 guide

arnaudgiuliani:
The two most-reported ViewModel resolution failures gave opaque errors and had no R8 guidance. This PR makes every failure mode self-explanatory and ships the missing R8 story. **No resolution behavior is removed** — these are diagnostics + a setup-error + docs.

Addresses #2044 and #2417.

## What was failing

1. **#2417(a)** — a ViewModel needing `SavedStateHandle` whose `CreationExtras` has no `SavedStateRegistryOwner` crashed with the raw androidx `IllegalArgumentException: CreationExtras must have a value by SAVED_STATE_REGISTRY_OWNER_KEY`.
2. **#2044** — when the parameter holder isn't on the stack (e.g. deferred `lazy { get() }`), `get<SavedStateHandle>()` fell through to the registry and threw `NoDefinitionFound` whose generic *"add a definition"* tail is wrong (you can't register a `SavedStateHandle`).
3. **#2417(b)** — the advised `viewModelScope { }` workaround then failed with `No definition found … on scope '_root_'`, because `viewModelScope { }` requires the `viewModelScopeFactory()` option to be resolvable.
4. The two viewmodel modules shipped **no R8 consumer rules**, and the R8 guide referenced by the other modules' `consumer-rules.pro` **didn't exist** — feeding the "minification breaks SavedStateHandle" confusion.

## Changes

**Actionable errors (code)**
- `AndroidParametersHolder` wraps the androidx crash in a Koin message pointing at the owner / `koinViewModel()` / constructor injection (#2417a).
- `CoreResolverV2.throwNoDefinitionFound` gives SavedStateHandle-specific guidance instead of the misleading "add a definition" tail (#2044), reusing the existing scope-targeting context. Matches on the type's final name segment so it works on JVM (FQN) **and** JS/WASM (simpleName) — #2312.
- `KoinViewModelFactory` detects a VM declared in `viewModelScope { }` while `viewModelScopeFactory()` is off (root resolution fails + VM registered under `ViewModelScopeArchetype`) and tells the user to enable the option (#2417b). Genuinely-missing VMs keep the normal message.

**R8 (docs + packaging)**
- `consumer-rules.pro` added to `koin-core-viewmodel` and `koin-compose-viewmodel` (`-dontwarn org.koin.**`), wired via `consumerProguardFiles` — verified embedded as `proguard.txt` in the release AARs.
- New `docs/reference/koin-android/r8-proguard.md`: Koin's resolution (`get<T>()` / `*Of`) is compile-time and **R8-safe** (no ViewModel keep-rules needed); what users *do* keep (Fragments, Workers, their saved-state classes); the SavedStateHandle-in-constructor rule; the `viewModelScope { }` + option requirement; and the JS/WASM/Native `qualifiedName` caveat.

## Why no "timing fix" for the intermittent #2044
`SavedStateHandle` is built from `CreationExtras` that only exist during `create()`, so deferred/lazy resolution is fundamentally unsupportable — that case now produces the clear #2417a error instead of a mystery. The intermittent direct-injection reports aren't deterministically reproducible; the improved messages localize *which* scope was targeted. We deliberately don't ship a speculative fix. (R8 is **not** the cause — the resolution path is R8-safe, see the guide.)

## Tests
`commonTest`, all falsified RED→GREEN on **JVM / wasmJs / native**:
- `SavedStateHandleErrorTest` — #2417a wrap + #2044 extras guidance
- `ViewModelScopeOptionTest` — #2417b missing-option error, with-option success, and not-over-firing for genuinely-missing VMs

## Verification
- Full `:core:koin-core:jvmTest` + `:core:koin-core-viewmodel:jvmTest` green; the KMP run caught the #2312 FQN/simpleName split (fixed)
- wasmJs + native: new tests green; only the documented pre-existing `createdAtStart` failures remain in koin-core
- `./gradlew apiCheck` — pass (internal changes only; no public API delta)
- Release AARs assemble and embed the consumer rules

## Scope
Diagnostics + #2417b setup-error + R8 docs. #2393 is by-design (documented separately). No version bump (no alpha will be published; ships in 4.2.2).


arnaudgiuliani:
**Scope update during review:** dropped the #2417(b) option-detection change (`KoinViewModelFactory` try/catch) — it deterministically regressed `ViewModelScopeArchetypeTest`'s #2387 SavedStateHandle property-get tests (base passes 3/3, branch failed 3/3; a *bare* rethrow also broke them, which means the VM resolution flow has exception-dependent control flow — likely androidx `ViewModelProvider` — that the wrapper perturbs). Not understood well enough to ship.

This PR now = the verified **1a/1b error improvements** (#2417a wrap + #2044 extras guidance) + **R8 consumer-rules & guide**. #2417(b) actionable-error stays open for a separate, properly-scoped fix with ViewModelScopeArchetypeTest as the guard.

PR #2446, opened 2026-06-12
Fix #2348 - ignore non-String environment properties (ClassCastException)

### LINKED ISSUE #2348: ClassCastException when reading environment properties
**Describe the bug**
When reading environment properties, Koin assumes that all properties are strings.
https://github.com/InsertKoinIO/koin/blob/461b5684684bb1b17411f27c35a955cdc90f299b/projects/core/koin-core/src/jvmMain/kotlin/org/koin/core/registry/PropertyRegistryExt.kt#L17

However, due to legacy java API weirdness, this is not guaranteed to be the case, and both keys and values could be arbitrary objects.

Therefore when Koin does the hard cast to strings, the application might crash.

**To Reproduce**
1. Set non-string properties via `System.setProperties()`
2. Start Koin framework

**Expected behavior**
Koin starts up fine, ignoring any invalid properties.

**Koin module and version:**
`koin-core-jvm:4.0.4`

```java
var properties = new Properties();

// save existing properties just in case
properties.putAll(System.getProperties());

// intentionally malformed input
properties.put(new Object(), new Object());

// just replace the existing map
System.setProperties(properties);

// start Koin.....
```

ref: https://github.com/Up-Mods/HT_DiscordBridge/issues/4
ref: https://github.com/Up-Mods/HT_DiscordBridge/issues/6

arnaudgiuliani:
## Problem
`PropertyRegistry.saveProperties` cast every `java.util.Properties` entry to `Map<String, String>`:
```kotlin
val propertiesMapValues = properties.toMap() as Map<String, String>
propertiesMapValues.forEach { (k: String, v: String) -> saveProperty(k, v) }
```
`Properties` can legally hold **non-String** keys/values (e.g. after `System.setProperties(...)` with arbitrary objects — happens with some libraries/runtimes). The destructuring cast then throws `ClassCastException` on `startKoin` / `loadEnvironmentProperties`. Fixes #2348.

## Fix
Filter to `String`/`String` entries and ignore the rest (debug-logged), per the issue's expected behavior ("start fine, ignoring invalid properties"). Removes the unchecked cast:
```kotlin
properties.forEach { (k, v) ->
    if (k is String && v is String) saveProperty(k, v)
    else _koin.logger.debug("ignore non-string property '$k'")
}
```

## Test
`EnvironmentPropertiesTest` (jvmTest) — a `Properties` containing a non-String key, a non-String value, and a both-non-String entry no longer throws; the valid string property still loads, the invalid ones are ignored. RED (ClassCastException) → GREEN.

## Verification
- `:core:koin-core:jvmTest` green; `apiCheck` pass.
- JVM-only change (`jvmMain`), so no off-JVM targets involved.

### DIFF
--- a/projects/core/koin-core/src/jvmMain/kotlin/org/koin/core/registry/PropertyRegistryExt.kt
+++ b/projects/core/koin-core/src/jvmMain/kotlin/org/koin/core/registry/PropertyRegistryExt.kt
@@ -10,13 +10,20 @@ import java.util.*
 /**
  *Save properties values into PropertyRegister
  */
-@Suppress("UNCHECKED_CAST")
 fun PropertyRegistry.saveProperties(properties: Properties) {
     _koin.logger.debug("load ${properties.size} properties")
 
-    val propertiesMapValues = properties.toMap() as Map<String, String>
-    propertiesMapValues.forEach { (k: String, v: String) ->
-        saveProperty(k, v)
+    // java.util.Properties can legally hold non-String keys/values (e.g. after
+    // System.setProperties(...) with arbitrary objects). The registry is keyed by
+    // String but stores Any values, so keep any String-keyed entry (value as-is)
+    // and only drop non-String keys — instead of hard-casting values to String and
+    // crashing (#2348).
+    properties.forEach { (k, v) ->
+        if (k is String && v != null) {
+            saveProperty(k, v)
+        } else {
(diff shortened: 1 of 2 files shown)


UpcraftLP:
Thank you for fixing this, now I can finally update my koin app.

Tho, is there any reason to drop all non-strings instead of allowing them as valid keys, like in #2358?

PR #2447, opened 2026-06-12
Fix #2386 - withOptions leaves stale index entries on qualifier/secondary change

### LINKED ISSUE #2386: `withOptions` silently creates stale index entries and inconsistent cross-property behavior
## Description

When using `withOptions` to mutate a `BeanDefinition`'s properties (`qualifier`, `secondaryTypes`, `_createdAtStart`) after initial registration, several undocumented and inconsistent behaviors occur:

1. Changing `qualifier` adds a new index entry but **never removes the old one**, creating a silent dual-index.
2. `secondaryTypes` are indexed using the **post-mutation** qualifier only, creating asymmetry with the stale primary index.
3. These properties interact in surprising ways when combined in a single `withOptions` block.

## Source Code Under Analysis

```kotlin
// org.koin.core.module.dsl.OptionDSL.kt
inline infix fun <T> KoinDefinition<T>.withOptions(
    options: DefinitionOptions<T>,
): KoinDefinition<T> {
    val def = factory.beanDefinition
    val primary = def.qualifier          // 1. snapshot old qualifier
    def.also(options)                    // 2. apply ALL mutations at once
    if (def.qualifier != primary) {
        module.indexPrimaryType(factory)  // 3. add new index (old NOT removed)
    }
    if (def.secondaryTypes.isNotEmpty()) {
        module.indexSecondaryTypes(factory) // 4. index secondary types with NEW qualifier
    }
    if (def._createdAtStart && factory is SingleInstanceFactory<*>) {
        module.prepareForCreationAtStart(factory) // 5. schedule eager creation
    }
    return this
}
```

Step 2 mutates `BeanDefinition` in-place, then steps 3-5 use the **mutated** state for indexing. But the **original** index entry from the initial `single()/factory()` call is never cleaned up from `Module.mappings`.

### How indexing works

```kotlin
// Module.kt
fun indexPrimaryType(instanceFactory: InstanceFactory<*>) {
    val def = instanceFactory.beanDefinition
    val mapping = indexKey(def.primaryType, def.qualifier, def.scopeQualifier)
    saveMapping(mapping, instanceFactory)  // adds to Module.mappings: LinkedHashMap<String, InstanceFactory>
}

// BeanDefinition.kt — index key is a deterministic string
fun indexKey(clazz: KClass<*>, typeQualifier: Qualifier?, scopeQualifier: Qualifier): String {
    return "${clazz.getFullName()}:${typeQualifier?.value ?: ""}:$scopeQualifier"
}
```

Resolution uses **direct O(1) map lookup**, not iteration:

```kotlin
// InstanceRegistry.kt
fun resolveDefinition(clazz: KClass<*>, qualifier: Qualifier?, scopeQualifier: Qualifier): InstanceFactory<*>? {
    val indexKey = indexKey(clazz, qualifier, scopeQualifier)
    return _instances[indexKey]
}
```

---

## Problem 1: Stale Qualifier Index

```kotlin
single<Foo>(named("a")) { Foo() }
    .withOptions { qualifier = named("b") }
```

`Module.mappings` after `withOptions`:

| Index Key | Factory | BeanDefinition.qualifier |
|-----------|---------|--------------------------|
| `"Foo:a:_root_"` | factory | `named("b")` |
| `"Foo:b:_root_"` | factory | `named("b")` |

Both entries point to the **same** `SingleInstanceFactory` object:

```kotlin
get<Foo>(named("a"))  // resolves via _instances["Foo:a:_root_"] → factory → works
get<Foo>(named("b"))  // resolves via _instances["Foo:b:_root_"] → factory → works
// Same factory, same singleton instance.
```

The definition's `qualifier` says `named("b")`, but the factory is still reachable via `named("a")` through the stale index entry.

### Chain: Multiple `withOptions` calls accumulate stale entries

```kotlin
single<Foo>(named("a")) { Foo() }
    .withOptions { qualifier = named("b") }  // primary was "a", now "b" → index "Foo:b:_root_" added
    .withOptions { qualifier = named("c") }  // primary was "b", now "c" → index "Foo:c:_root_" added
```

Now **three** index entries in `Module.mappings` all point to the same factory:
- `"Foo:a:_root_"` (stale from initial `single()`)
- `"Foo:b:_root_"` (stale from first `withOptions`)
- `"Foo:c:_root_"` (current from second `withOptions`)

---

## Problem 2: `secondaryTypes` Indexed Inconsistently with Stale Qualifier

`indexSecondaryTypes` uses the **current** `def.qualifier` (post-mutation), but the stale primary index still uses the **old** qualifier. This creates an asymmetry where the same definition's primary type and secondary types have different reachability:

```kotlin
single<Foo>(named("a")) { FooImpl() }
    .withOptions {
        qualifier = named("b")
        secondaryTypes = listOf(Bar::class)
    }
```

`Module.mappings`:

| Index Key | Factory |
|-----------|---------|
| `"Foo:a:_root_"` | factory (stale, from initial `single()`) |
| `"Foo:b:_root_"` | factory (from `indexPrimaryType` with new qualifier) |
| `"Bar:b:_root_"` | factory (from `indexSecondaryTypes` with new qualifier) |

Resolution:
```kotlin
get<Foo>(named("a"))  // OK   — stale primary index
get<Foo>(named("b"))  // OK   — new primary index
get<Bar>(named("b"))  // OK   — secondary type indexed with new qualifier
get<Bar>(named("a"))  // FAIL — secondary type NOT indexed under old qualifier
```

**The asymmetry**: `Foo` is reachable via both `named("a")` and `named("b")`, but `Bar` is only reachable via `named("b")`. They are the same definition — a user would expect consistent reachability.

### Variant: qualifier = null + secondaryTypes

```kotlin
single<Foo>(named("a")) { FooImpl() }
    .withOptions {
        qualifier = null
        secondaryTypes = listOf(Bar::class)
    }
```

| Index Key | Factory |
|-----------|---------|
| `"Foo:a:_root_"` | factory (stale) |
| `"Foo::_root_"` | factory (new, unqualified) |
| `"Bar::_root_"` | factory (secondary, unqualified) |

```kotlin
get<Foo>()            // OK   — new unqualified index
get<Foo>(named("a"))  // OK   — stale index
get<Bar>()            // OK   — secondary type, unqualified
get<Bar>(named("a"))  // FAIL — asymmetry!
```

---

## Problem 3: `_createdAtStart` with Qualifier Change

```kotlin
single<Foo>(named("a")) { println("created"); Foo() }
    .withOptions {
        qualifier = named("b")
        _createdAtStart = true
    }
```

`prepareForCreationAtStart` adds the factory to `Module.eagerInstances`. At Koin startup, `createEagerInstances()` calls `factory.get(defaultContext)`, which creates and caches the singleton in the `SingleInstanceFactory`.

After startup:
- `get<Foo>(named("b"))` → returns cached instance (expected)
- `get<Foo>(named("a"))` → same factory via stale index → returns same cached instance (unexpected)

The instance was eagerly created for `named("b")`, but is silently accessible via `named("a")`.

---

## Problem 4: Conflict with Explicit Definitions

```kotlin
module {
    single<Foo>(named("a")) { Foo("aliased") }    // index: "Foo:a:_root_" → factory1
        .withOptions { qualifier = null }           // index: "Foo::_root_"  → factory1

    single<Foo> { Foo("explicit-default") }         // index: "Foo::_root_"  → factory2 (overwrites factory1)
}
```

`Module.mappings` is a `LinkedHashMap` — same key means last-write-wins deterministically:

| Index Key | Factory |
|-----------|---------|
| `"Foo:a:_root_"` | factory1 (stale, still points to the first definition) |
| `"Foo::_root_"` | factory2 (overwritten by explicit `single<Foo>`) |

```kotlin
get<Foo>()            // factory2 → Foo("explicit-default")
get<Foo>(named("a"))  // factory1 → Foo("aliased")  — stale index, different instance!
```

Two `get` calls for the "same" type return **different instances** from **different factories**. The stale index silently kept the old factory alive.

---

## Summary Table

| Scenario | Expected | Actual |
|----------|----------|--------|
| Change qualifier | Old index removed, new index added | Both indexes exist (stale + new) |
| qualifier + secondaryTypes | Secondary types indexed under all reachable qualifiers | Secondary types only indexed under new qualifier — asymmetric reachability |
| Multiple `withOptions` calls | Each call replaces the previous qualifier | Index entries accumulate with each call |
| qualifier + `_createdAtStart` | Eager instance only reachable via final qualifier | Also reachable via stale qualifier |
| Conflict with explicit definition | Explicit definition fully replaces aliased one | Stale index keeps old factory alive alongside new one |

## Suggestion

**Option A: Clean up stale indexes** — When `withOptions` detects a qualifier change, remove the old index entry from `Module.mappings` before adding the new one. This makes `withOptions` a true "replace" operation.

**Option B: Document as alias mechanism** — If dual-indexing is intentional, document it explicitly and ensure `secondaryTypes` are indexed under **all** reachable qualifiers (both old and new) for consistency.

**Option C: Provide an explicit alias API** — Separate "change qualifier" from "add alias":
```kotlin
// Replace qualifier (removes old index)
.withOptions { qualifier = named("b") }

// Add alias (keeps old index, adds new one, indexes secondaryTypes under both)
.alsoQualifiedAs(null)
```

## Environment

- Koin version: 4.1.1
- Kotlin: 2.1.x
- Platform: KMP (JVM / iOS / Desktop)

arnaudgiuliani:
## Problem
`withOptions` mutates a `BeanDefinition` in place, then re-indexes from the mutated state — but never removes the entry registered under the **previous** coordinates. So:
```kotlin
single<Foo>(named("a")) { Foo() } withOptions { qualifier = named("b") }
```
leaves **both** `Foo:a` (stale) and `Foo:b` in `Module.mappings`, pointing at the same factory — `get<Foo>(named("a"))` still resolves even though the definition's qualifier is now `b`. With `secondaryTypes` it gets worse: secondaries are indexed only under the new qualifier, so `Foo` is reachable via old+new but `Bar` only via new (asymmetry). Fixes #2386.

## Fix
Snapshot the previous `qualifier` + `secondaryTypes` before applying the options; when either changes, drop the stale `mappings` entries (old primary + old secondaries under the old qualifier) before re-indexing with the new coordinates:
```kotlin
val previousQualifier = def.qualifier
val previousSecondaryTypes = def.secondaryTypes.toList()
def.also(options)
if (def.qualifier != previousQualifier || def.secondaryTypes != previousSecondaryTypes) {
    module.mappings.remove(indexKey(def.primaryType, previousQualifier, def.scopeQualifier))
    previousSecondaryTypes.forEach { module.mappings.remove(indexKey(it, previousQualifier, def.scopeQualifier)) }
    module.indexPrimaryType(factory)
    if (def.secondaryTypes.isNotEmpty()) module.indexSecondaryTypes(factory)
}
```
No-op when nothing index-relevant changed.

## Behavioral note (resolution semantics)
A definition is **no longer reachable under a qualifier it had before `withOptions` changed it** — this removes unintended/undocumented reachability (the dual-index). Resolution by the definition's *current* qualifier is unchanged.

## Tests
- `WithOptionsStaleIndexTest` (commonTest, RED→GREEN): qualifier change drops the stale primary index; primary + secondary reachability stay consistent.
- `NewDSLTest`: corrected three `mappings.size` expectations that were **counting the stale entry** (3→2, 3→2, 4→3) — they encoded the bug.

## Verification
- `:core:koin-core:jvmTest` green; wasmJs green except the documented pre-existing `createdAtStart` reds; native validated via JVM+wasmJs (native run hits the known pre-existing `ServiceMessagesParser` backtick-reporter crash, unrelated).
- `./gradlew apiCheck` — pass (no public API change; `mappings` is `@KoinInternalApi`).

### DIFF
--- a/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/module/dsl/OptionDSL.kt
+++ b/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/module/dsl/OptionDSL.kt
@@ -7,6 +7,7 @@ import org.koin.core.definition.BeanDefinition
 import org.koin.core.definition.Callbacks
 import org.koin.core.definition.KoinDefinition
 import org.koin.core.definition.OnCloseCallback
+import org.koin.core.definition.indexKey
 import org.koin.core.instance.SingleInstanceFactory
 import org.koin.core.module.OptionDslMarker
 import org.koin.core.qualifier.StringQualifier
@@ -41,13 +42,22 @@ inline infix fun <T> KoinDefinition<T>.withOptions(
     options: DefinitionOptions<T>,
 ): KoinDefinition<T> {
     val def = factory.beanDefinition
-    val primary = def.qualifier
+    val previousQualifier = def.qualifier
+    val previousSecondaryTypes = def.secondaryTypes.toList()
     def.also(options)
-    if (def.qualifier != primary) {
+    // #2386: mutating qualifier/secondaryTypes changes the definition's index coordinates. The
+    // original entries (from single()/factory() or a prior withOptions) must be removed before
+    // re-indexing, otherwise the definition stays reachable under its old qualifier and its
+    // secondary types desync from the primary index.
+    if (def.qualifier != previousQualifier || def.secondaryTypes != previousSecondaryTypes) {
+        module.mappings.remove(indexKey(def.primaryType, previousQualifier, def.scopeQualifier))
+        previousSecondaryTypes.forEach { secondaryType ->
+            module.mappings.remove(indexKey(secondaryType, previousQualifier, def.scopeQualifier))
+        }
         module.indexPrimaryType(factory)
-    }
-    if (def.secondaryTypes.isNotEmpty()) {
-        module.indexSecondaryTypes(factory)
+        if (def.secondaryTypes.isNotEmpty()) {
+            module.indexSecondaryTypes(factory)
+        }
     }
     if (def._createdAtStart && factory is SingleInstanceFactory<*>) {
         module.prepareForCreationAtStart(factory)
--- a/projects/core/koin-core/src/commonTest/kotlin/org/koin/core/WithOptionsStaleIndexTest.kt
+++ b/projects/core/koin-core/src/commonTest/kotlin/org/koin/core/WithOptionsStaleIndexTest.kt
@@ -1,0 +1,57 @@
+package org.koin.core
+
+import org.koin.Simple
+import org.koin.core.module.dsl.withOptions
+import org.koin.core.qualifier.StringQualifier
+import org.koin.core.qualifier.named
+import org.koin.dsl.koinApplication
+import org.koin.dsl.module
+import kotlin.test.Test
+import kotlin.test.assertNotNull
+import kotlin.test.assertNull
+
+/**
+ * Regression test for #2386. Mutating a definition's qualifier/secondaryTypes via withOptions
+ * added new index entries but never removed the originals, leaving stale dual-index entries:
+ * the definition stayed reachable under its OLD qualifier, and secondary types were indexed
+ * only under the new qualifier (asymmetry). After the fix, the old coordinates must be dropped.
+ */
+class WithOptionsStaleIndexTest {
+
+    @Test
+    fun qualifier_change_drops_stale_primary_index_2386() {
+        val koin = koinApplication {
+            modules(
+                module {
+                    single(named("a")) { Simple.ComponentA() } withOptions { qualifier = StringQualifier("b") }
+                },
+            )
+        }.koin
+
+        // reachable under the new qualifier
+        assertNotNull(koin.getOrNull<Simple.ComponentA>(named("b")))
+        // NOT reachable under the old qualifier (stale index removed)
+        assertNull(koin.getOrNull<Simple.ComponentA>(named("a")))
+    }
+
+    @Test
+    fun qualifier_change_keeps_primary_and_secondary_reachability_consistent_2386() {
+        val koin = koinApplication {
+            modules(
+                module {
+                    single(named("a")) { Simple.Component1() } withOptions {
+                        qualifier = StringQualifier("b")
+                        secondaryTypes = listOf(Simple.ComponentInterface1::class)
+                    }
+                },
+            )
+        }.koin
+
+        // primary + secondary both reachable under the NEW qualifier
+        assertNotNull(koin.getOrNull<Simple.Component1>(named("b")))
+        assertNotNull(koin.getOrNull<Simple.ComponentInterface1>(named("b")))
+        // neither reachable under the OLD qualifier — consistent (no asymmetry)
+        assertNull(koin.getOrNull<Simple.Component1>(named("a")))
+        assertNull(koin.getOrNull<Simple.ComponentInterface1>(named("a")))
+    }
+}
--- a/projects/core/koin-core/src/commonTest/kotlin/org/koin/dsl/NewDSLTest.kt
+++ b/projects/core/koin-core/src/commonTest/kotlin/org/koin/dsl/NewDSLTest.kt
@@ -30,7 +30,8 @@ class NewDSLTest : KoinCoreTest() {
                 bind<IClassA>()
             }
         }
-        assertEquals(3, module.mappings.size)
+        // qualified primary + secondary; the original unqualified primary index is dropped (#2386)
+        assertEquals(2, module.mappings.size)
         val factory = module.mappings.values.iterator().next()
         assertTrue {
             factory is SingleInstanceFactory<*> && factory.beanDefinition._createdAtStart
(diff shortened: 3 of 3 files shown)


arnaudgiuliani:
⚠️ Converting to draft — this fix regresses `koin-android` `DSLExtendedTest.android dsl`.

**Root cause of the regression:** with two same-type definitions —
```kotlin
viewModelOf(::MyViewModel)               // (1) indexes MyViewModel:null
viewModelOf(::MyViewModel){ named("bis") } // (2) indexes MyViewModel:null at creation (overwrites (1)!), then withOptions moves it to :bis
```
— the `*Of(ctor){options}` builders index under the **null** qualifier first and only change it afterward, so (2) transiently collides with (1) at `MyViewModel:null`. On `main` the leftover stale `:null` entry is what keeps `getOrNull<MyViewModel>()` non-null; removing it (the #2386 fix) exposes that (1) was already clobbered → unqualified resolves to null.

**Implication:** the clean fix for #2386's single-definition cases is correct (verified by `WithOptionsStaleIndexTest`), but a complete fix must also stop `*Of(ctor){options}` from indexing under the pre-options qualifier — i.e. apply options **before** the initial `indexPrimaryType`. That's a broader change to the `*Of` builders (Factory/Single/Scoped/ViewModelOf). Holding until that's designed.

arnaudgiuliani:
Closing — #2386 is deferred to **4.3.0** (needs the deeper `*Of`-builder indexing-order fix, not a contained patch change). Investigation, root cause, and the single-definition fix + `WithOptionsStaleIndexTest` are recorded here and on #2386 for whoever picks it up.

PR #2464, opened 2026-08-24
Fix #2463 - close Ktor request scope when the call fails

### LINKED ISSUE #2463: Ktor request scope remains open and registered when downstream response completion fails
**Describe the bug**

`koin-ktor` creates a request scope during `CallSetup` and closes it only from its `ResponseSent` handler.

Current implementation:

https://github.com/InsertKoinIO/koin/blob/4.2.2/projects/ktor/koin-ktor/src/commonMain/kotlin/org/koin/ktor/plugin/KoinPlugin.kt

```kotlin
on(CallSetup) { call ->
}

on(ResponseSent) { call ->
    call.attributes[KOIN_SCOPE_ATTRIBUTE_KEY].close()
}
```

Ktor's `ResponseSent` hook calls `proceed()` before invoking its handler. If a later response hook or downstream send operation throws, execution never returns to Koin's handler and `Scope.close()` is skipped.

`RequestScope` stores the `ApplicationCall` as its source:

https://github.com/InsertKoinIO/koin/blob/4.2.2/projects/ktor/koin-ktor/src/commonMain/kotlin/org/koin/ktor/plugin/RequestScope.kt

The resulting scope remains open and registered in `ScopeRegistry`, retaining the complete `ApplicationCall` graph.

**Observed impact**

We found this while analyzing a 1+ GB heap from a long-running Ktor service.

The heap contained approximately 18,000 Koin request scopes. Their GC-root path was:

```text
Koin
  -> ScopeRegistry
  -> request Scope
  -> ApplicationCall
```

The retained calls included HTTP headers, cookies, coroutine continuations, request models, and request-scoped dependencies.

**To Reproduce**

I added the following test to `KoinPluginRunTest` on current Koin `main`/4.2.2:

```kotlin
@Test
fun `closes request scope when a later response completion hook fails`() = testApplication {
    lateinit var requestScope: Scope
    val responseCompletionHookRan = AtomicBoolean()

    application {
        install(Koin)

        // A hook installed later runs first after proceed() in the
        // ResponseSent phase.
        install(
            createApplicationPlugin(
                name = "FailResponseCompletionBeforeKoinCleanup",
            ) {
                on(ResponseSent) {
                    responseCompletionHookRan.set(true)
                    error("response completion failed before Koin cleanup")
                }
            },
        )

        routing {
            get("/scope") {
                requestScope = call.scope
                call.respond(HttpStatusCode.OK)
            }
        }
    }

    runCatching {
        client.get("/scope")
    }

    assertTrue(
        responseCompletionHookRan.get(),
        "response completion failure hook should run",
    )

    val scopeClosed = requestScope.closed
    val scopeStillRegistered =
        requestScope.getKoin().getScopeOrNull(requestScope.id) != null

    assertTrue(
        scopeClosed && !scopeStillRegistered,
        "Koin-created request scope should be closed and removed " +
            "(closed=$scopeClosed, registered=$scopeStillRegistered)",
    )
}
```

Run it with:

```shell
./gradlew :ktor:koin-ktor:jvmTest \
  --tests 'org.koin.ktor.ext.KoinPluginRunTest.closes request scope when a later response completion hook fails'
```

The test compiles and fails with:

```text
Koin-created request scope should be closed and removed
(closed=false, registered=true)
```

The preceding assertion confirms that the injected failure hook ran. The final assertion directly observes that the Koin-created scope remains both open and registered.

**Expected behavior**

The request scope should close and be removed from `ScopeRegistry` whenever the Ktor call terminates, including when:

- response processing succeeds
- response processing throws
- response sending fails
- the call is cancelled

Cleanup should remain safe if Koin's normal `ResponseSent` handler has already closed the scope.

**Koin module and version**

- `koin-ktor:4.2.2`
- Ktor 3.4.0
- JVM 21
- Reproduced on commit `dc86ef8dd8fbe8564fb7453c03f5b738da3450bb`

**Possible fix**

Request-scope cleanup could use a lifecycle path that runs after successful completion, failure, and cancellation. For example, Koin could combine its existing `ResponseSent` handling with failed-call cleanup or wrap downstream call processing in `try/finally`.

The cleanup should use the scope already stored in `KOIN_SCOPE_ATTRIBUTE_KEY` and avoid closing it twice.

inemtsev:
The Koin Ktor plugin creates a request scope in CallSetup and closes it in ResponseSent. ResponseSent alone is not enough for a call that fails, so the scope stays registered in ScopeRegistry and keeps the whole ApplicationCall graph alive. Two paths leak:

- A ResponseSent interceptor installed after Koin runs before Koin's. If it throws, the exception unwinds past Koin's handler and the scope never closes.
- A route handler that throws before responding is answered by the engine fallback, which never enters the send pipeline, so ResponseSent never fires at all.

This PR adds an on(CallFailed) hook that closes the scope on the failure path and moves cleanup into a shared closeRequestScope() that tolerates a missing scope attribute and a double close (StatusPages can respond after a failure, which fires both hooks).

KoinPluginRunTest gets three new tests. Each one grabs the request scope inside a route handler, lets the call finish, and then asserts that the scope is closed and no longer registered in ScopeRegistry.

- closes request scope on normal response completion: the handler responds 200 OK. This path already worked on main; the test is there to prove the change did not break it.
- closes request scope when a later response completion hook fails: a second plugin installed after Koin registers its own ResponseSent hook that throws. Because it was installed later, it runs before Koin's hook, so Koin's cleanup never ran. Fails on main, passes here.
- closes request scope when route handler fails before responding: the handler throws without calling respond. The engine answers with its own error response and never enters the send pipeline, so ResponseSent never fires. Fails on main, passes here.

Fixes #2463


inemtsev:
Hi @arnaudgiuliani ,

Can you please take a look? This can prevent large memory usage during error spikes.

kchinburarat:
vote up here

Issue #2252, opened 2025-07-17
Support Version Catalog


zhangz1han:
Many libraries with multiple artifacts already support Version Catalog to simplify dependency management. It does not need to declare each library and version in `libs.versions.toml`, which is simpler and more convenient than BOM.

### Example
- [Ktor Version Catalog](https://ktor.io/docs/whats-new-320.html#published-version-catalog)
- [AndroidX Version Catalog](https://maven.google.com/web/index.html?q=gradle-version-catalog)
- [KotlinCrypto Version Catalog](https://github.com/KotlinCrypto/version-catalog)

### Usage
```kts
// settings.gradle.kts
dependencyResolutionManagement {
    versionCatalogs {
        create("koinLibs") {
            from("io.insert-koin:koin-version-catalog:4.1.0")
        }
    }
}

// <module>/build.gradle.kts
dependencies {
    implementation(koinLibs.core)
    testImplementation(koinLibs.test)
    // ...
}
```

kibettheophilus:
@zhangzih4n thanks for the suggestion, will pick it up

arnaudgiuliani:
yes, good proposal 👍

arnaudgiuliani:
@kibettheophilus you can open a PR if you want. I can still ad it for 4.2

hmzgtl16:
Are there any active PRs for this yet, or may I take it over for the 4.3.0 release?

kibettheophilus:
@hmzgtl16 let e open a PR

arnaudgiuliani:
need to re-align on main. I will open 4.3 branch soon 👍

PR #2466, opened 2026-09-07
Setup koin version catalog

kibettheophilus:
#2252


kibettheophilus:
closed in favor of #2465

